PR #9, opened 2016-05-31
Remove unneeded unit tests for unneeded UUID method

### LINKED ISSUE #8: Remove patch 10: Clean up and test the UUID code patch
[Patch 10](https://github.com/WaterfallMC/Waterfall/blob/master/BungeeCord-Patches/0010-Clean-up-and-unit-test-the-UUID-code.patch) contains a better replacement for `replaceAll`, but I think adding a method to the utility class and adding unit tests is a little much for removing the word `All` twice. There's no need to unit test a standard library method, that's Oracle's job. I think a patch to change `replaceAll` to `replace` is valid, but the rest is overkill.

phase:
getMojangUUID was a replacement for the regex conversion of UUIDs to Mojang's format. I have removed this method and the associated unit tests because it is nothing more than a wrapper around String#replace. This new patch simply replaces the call to String#replaceAll with a call to String#replace. Patch 0011 was affected by this change, and has been fixed.

This closes #8.


minecrafter:
We made it a separate method in case we wanted to perform further changes to it.

Now that I think of it... `replace(String, String)` still uses the regex library in Java.

phase:
It doesn't have to compile the arguments as regex, which should speed up something.

Janmm14:
See patch 11, where it stops being a wrapper.

PR #15, opened 2016-06-06
Add 0034-Fix-Cannot-get-ID-for-packet-class-Kick.patch

kamcio96:
Propably `disconnecting` isn't required.

I didn't test it :wink:
Jar file: http://kamcio96.pl/Waterfall.jar


Techcable:
Wait what?
I don't get why this is nessicary.

kamcio96:
See IRC

Techcable:
I think we should try and figure out exactly which state is triggering the issue ;)
Thanks for the suggestion though :)

PR #37, opened 2016-06-29
Added dynamic server addition/removal api methods

fuzzybot:


Techcable:
Can you put this in a seperate patch file?

PR #42, opened 2016-06-29
Since servers is mutable, lock on another Object: serversLock

Black-Hole:
It might even be null.


BungeeCord-Patches/0037-Add-dynamic-server-addition-removal-api.patch
```diff
@@ -187,15 +187,23 @@ index db9ebbd..112420a 100644
  
  import net.md_5.bungee.BungeeCord;
  import net.md_5.bungee.api.Favicon;
-@@ -61,6 +63,7 @@ public abstract class Configuration implements ProxyConfig
+@@ -42,6 +44,7 @@ public abstract class Configuration implements ProxyConfig
+      * Set of all listeners.
+      */
+     private Collection<ListenerInfo> listeners;
++    private Object serversLock = new Object(); // Waterfall
```
Techcable: This should be final

Black-Hole: done

PR #139, opened 2017-05-27
Use Log4j for logging

Foorack:
Default log file is logs/proxy.log with rollover daily and on startup. Console and file output format is the same.

- All Log4j loggers are asynchronous
- Supports Loggers created through j.u.l.Logger.getLogger()

Credits goes to zreed!
https://github.com/zreed/BungeeCord/commit/3c8cba3ce9c7953b73fed7af3d7ef89a2a322353

This patch should be removed if it is decided that the pull request in BungeeCord should be merged. There are many advantages to using log4j, please read the parent pull request for more information.

### DIFF
--- a/BungeeCord-Patches/0042-Use-Log4j-for-logging.patch
+++ b/BungeeCord-Patches/0042-Use-Log4j-for-logging.patch
@@ -1,0 +1,786 @@
+From 1814235858fca5ec335ba48e0071e21a688b03c9 Mon Sep 17 00:00:00 2001
+From: Foorack <maxfaxalv@gmail.com>
+Date: Sat, 27 May 2017 12:24:37 +0200
+Subject: [PATCH] Use Log4j for logging Default log file is logs/proxy.log with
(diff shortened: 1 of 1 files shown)


Techcable:
I think we should use slf4j for logging, so we're more insulated from the underlying implementation. Do you have any objections to that @zreed ?

zreed:
I went with Log4j mainly due to its use in Minecraft (feature parity) and its more efficient translation of j.u.l logging. Why not just also add slf4j-api to the API & slf4j->log4j2 to the Proxy?

Techcable:
I will, just wanted your feedback. Thanks for helping improve our logging situation!

Foorack:
Hello. :) Any status update on this? Can it be merged?

Foorack:
@Techcable Ping. This nearly 3 months old now.

jamierocks:
No longer necessary, having merged https://github.com/WaterfallMC/Waterfall/pull/176.

PR #161, opened 2017-07-26
Make usage of the Netty DNS resolver configurable

natemort:
While it is generally a better idea to use the asynchronous DNS resolver and not have to worry about managing our own cache, we've been unable to resolve the session servers since the update to Netty 4.1.13.  Looking through their [commit history](https://github.com/netty/netty/commits/4.1/resolver-dns) there's a number of significant behavioral changes to their DNS resolution that were introduced with 4.1.13, with more still being made for their next release.  [Issues have also been reported in the past ](https://github.com/WaterfallMC/Waterfall/issues/117) that are likely caused by Netty's DNS resolver as well.

We had no issues with previous versions of Waterfall, and when running either this patch or a version of Waterfall with the Netty DNS resolver removed are able to resolve the session servers.

I'm not sure exactly what you're looking for in terms of patches, so I just went with modifying the existing patch.  I'm of course open to any changes as I'd rather avoid managing my own fork.


jamierocks:
Absolutely, I suspected that the async DNS resolver was down to some recent issue our users have been having. Regardless, it is probably a good idea to have this be configurable. Have you tested these changes?

natemort:
Yeah, we have no issues on our network with the default DNS resolver, and are completely unable to resolve sessionserver.mojang.com when using the Netty resolver.  I've tested this exact patch and verified that it works for both enabling and disabling Netty DNS.

jamierocks:
In that case, let's get this merged! Cheers ;)

PR #176, opened 2017-09-22
Use Log4j, add console improvements and allow plugins to use SLF4 for logging

stephan-gh:
This is a combination of moving Waterfall to Log4J (related to #139), additional console improvements using TerminalConsoleAppender (see https://github.com/PaperMC/Paper/pull/728) and allowing plugins to use SLF4J (related to https://github.com/PaperMC/Paper/pull/890).

It's similar to #139 but implemented differently. The only thing both PRs have in common is that Log4J2 is used (together with its async logging functionality). Otherwise they are different:

- [TerminalConsoleAppender](https://github.com/Minecrell/TerminalConsoleAppender) and JLine 3 are used for console improvements (e.g. colors, command completion, ...). See https://github.com/PaperMC/Paper/pull/728 for new features introduced by switching.
- Logged messages are forwarded using a custom `Handler` instead of Log4J's custom JUL `LogManager`. This is mostly because Log4J's custom `LogManager` prevents use of custom handlers attached to the JUL `Logger` and some plugins may make use of that. It's now similar to what Spigot does to forward them and I think it has shown to work quite well.
- `System.out` and `System.err` are redirected to the logger using Log4J IOStreams

There are two additional commits that are equivalent to the changes in https://github.com/PaperMC/Paper/pull/890. Plugin prefixes are handled using logger names in the Log4J configuration and plugins may now use SLF4J for logging.

As far I can tell, everything is already working correctly but I still need to do some cleanup...

### DIFF
--- a/BungeeCord-Patches/0044-Use-Log4j2-for-logging-and-TerminalConsoleAppender-f.patch
+++ b/BungeeCord-Patches/0044-Use-Log4j2-for-logging-and-TerminalConsoleAppender-f.patch
@@ -1,0 +1,529 @@
+From 51290fa75163f0f8ddcb508aa0b0aa20ffb00db1 Mon Sep 17 00:00:00 2001
+From: Minecrell <dev@minecrell.net>
+Date: Fri, 22 Sep 2017 12:46:47 +0200
+Subject: [PATCH] Use Log4j2 for logging and TerminalConsoleAppender for
+ console
+
+
+diff --git a/bootstrap/pom.xml b/bootstrap/pom.xml
+index 7bc2dff7..6e8dd81e 100644
+--- a/bootstrap/pom.xml
++++ b/bootstrap/pom.xml
+@@ -80,7 +80,17 @@
+                             </excludes>
+                         </filter>
+                     </filters>
++                    <transformers>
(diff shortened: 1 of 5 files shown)


stephan-gh:
As discussed on IRC, I've added a Log4j configuration file that replicates the old BungeeCord setup. It can be enabled by running Waterfall using `-Dlog4j.configurationFile=log4j2-bungee.xml`. However, the default is a setup that is very similar to what Vanilla/Paper/Sponge/Forge use.

BungeeCord-Patches/0044-Use-Log4j2-for-logging-and-TerminalConsoleAppender-f.patch
```diff
@@ -0,0 +1,529 @@
...
++        /*
+         while ( bungee.isRunning )
+         {
+             String line = bungee.getConsoleReader().readLine( ">" );
+@@ -32,6 +35,7 @@ public class Test
+                     bungee.getConsole().sendMessage( ChatColor.RED + "Command not found" );
+                 }
+             }
+-        }
++        }*/
```
jamierocks, line 295: `*/` should be on the next line

kashike (approved):

kashike:
Thanks, @Minecrell.

jamierocks (approved):
Stuff to be done while merging

PR #187, opened 2017-11-05
Fix #178 Error decoding packet unknown with contents

### LINKED ISSUE #178: Error decoding packet unknown with contents
I use latest build Waterfall (#147) and some time ago began to receive such error

Latest build paper spigot 1.12 (with latest ViaVersion 1.3.0)
https://pastebin.com/SVn8qBuG

Latest build paper spigot 1.8.8 (with latest ViaVersion 1.3.0)
https://pastebin.com/k0Ge5qVc

Polyana-Fontes:
It was happening on my server too and this changed fixed the issue.

This is my first time pushing changes to a patch-based repository, let me know if I did anything wrong generating this patch.

I only added:
```java
            if ( in.readerIndex() == in.writerIndex() )
            {
                return;
            }
```

to the MinecraftDecoder.decode() method as the error happens when the ByteBuf has no byte available to be read.

### DIFF
--- a/BungeeCord-Patches/0003-Add-Waterfall-configuration-files.patch
+++ b/BungeeCord-Patches/0003-Add-Waterfall-configuration-files.patch
@@ -1,11 +1,11 @@
-From 1893ed633fc97a855d0e657baa09feec9f007d11 Mon Sep 17 00:00:00 2001
+From db1bf650eabb7d9056ed89677c737ef455f6a4d1 Mon Sep 17 00:00:00 2001
 From: Techcable <Techcable@techcable.net>
 Date: Tue, 25 Oct 2016 11:58:37 -0400
 Subject: [PATCH] Add Waterfall configuration files
 
 
 diff --git a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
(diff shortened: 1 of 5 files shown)


MarkTwoFive:
If the error occurrs when there's no byte available to read wouldn't ``InputStream#available`` be a better way to determine when to exit early?

astei:
@MarkL4YG Netty doesn't use `InputStream`s, it uses `ByteBuf`s.

jamierocks:
Can anyone confirm that this resolves #178?

PurpleIsEverything:
Tested this recently with no negative effects it remains to be seen if it's actually doing anything.

Fejm:
Tested & working

PurpleIsEverything:
@jamierocks Please merge this it works totally fine.

electronicboy:
I'm pretty sure I stated this on IRC a while back, The issue here is that something is sending/throwing out invalid data somewhere. This "works" as a means for bodging around the issue, but isn't a fix whatsoever

jamierocks:
Having looked into this more closely a while back, I'm inclined to agree on that. However, I would like to avoid kicking players.

MarkTwoFive:
Sounds reasonable enough. I would suggest keeping the issue open labeling it something like "work-around deployed" after merging so it doesn't get lost just because it's suppressed

PR #200, opened 2018-01-19
Provide option to turn off connection reset messages from ServerInfo#…

Foorack:
…ping
Fixes https://github.com/WaterfallMC/Waterfall/issues/197

### DIFF
--- a/BungeeCord-Patches/0046-Provide-option-to-turn-of-connection-reset-messages-.patch
+++ b/BungeeCord-Patches/0046-Provide-option-to-turn-of-connection-reset-messages-.patch
@@ -1,0 +1,91 @@
+From 8d89a3929172f23b27988c4b03e87f8fe3333fec Mon Sep 17 00:00:00 2001
+From: Foorack <maxfaxalv@gmail.com>
+Date: Fri, 19 Jan 2018 23:29:44 +0100
+Subject: [PATCH] Provide option to turn of connection reset messages from
+ ServerInfo#ping
+
+
+diff --git a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+index 43519ce1..af2a4a52 100644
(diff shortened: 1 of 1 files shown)


jamierocks:
The warning message is totally justified, and as I've mentioned previously there already exists a way to omit them (through use of a custom log4j2 configuration). I don't know why BTLP increases the frequency of the warnings, or whether it causes them. If it does, I would say that this is not a fix - and that BTLP is broken (depending on whether it works on Bungee of course).

In regards to the actual pr, would you mind explaining: `!( handler instanceof PingHandler ) `?

Foorack:
@jamierocks @CodeCrafter47 I do not think it in this case is justified. BTLP is not broken. It uses the API function ServerInfo#ping to check if a server is online or not, which I think should be a valid use case of that function. The API then returns the answer to a callback. However, if the server is offline then bungee will also print an error message which is not possible to prevent or turn off. Simply pinging a server should not produce a warning message.

The reason why it increases when using BungeeTabListPlus is because it regularly calls `ServerInfo#ping` at a fixed interval to check if the backend servers are online. The BungeeCord project closed a similar issue report in denial and claimed it was simply a network issue or plugin misconfiguration (which is not the case).
https://github.com/SpigotMC/BungeeCord/issues/1959

This would surely be likely to circumvent with a custom log4j2 configuration file however passing around those config files would be inconvenient, especially when this affects any user who uses any plugin which takes use of ServerInfo#ping.

With regards to the pr, the intended logic was to always print the error like normal if it is not a PingHandler or if it is and the config value is true. :smiley:

jamierocks:
I'm going to close this pr for the time being, based on the following.
1. This special-case is wrongly positioned, and does not fix the issue
2. I'd sooner see a special-casing effort go through Bungeecord first, Waterfall second

Foorack:
For future reference, this issue has now been resolved upstream as of commit https://github.com/SpigotMC/BungeeCord/commit/e690a7b38908df542a4950914eaa6e6a68fb5d97

PR #206, opened 2018-02-08
Make InitialHandler connections logging optional

sgdc3:
I know it can be done with logger filters but i noticed some performance issues when my network is under heavy bot attack. (100 lines of log per second)

### DIFF
--- a/BungeeCord-Patches/0046-Make-InitialHandler-connections-logging-optional-dis.patch
+++ b/BungeeCord-Patches/0046-Make-InitialHandler-connections-logging-optional-dis.patch
@@ -1,0 +1,77 @@
+From d8b5c0e6f8a30d61669ae4cd7bd4a675d6bfe1a9 Mon Sep 17 00:00:00 2001
+From: Gabriele C <sgdc3.mail@gmail.com>
+Date: Thu, 8 Feb 2018 19:10:52 +0100
+Subject: [PATCH] Make InitialHandler connections logging optional (disabled by
+ default)
+
+
+diff --git a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+index 43519ce1..2631d6bb 100644
+--- a/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
++++ b/api/src/main/java/net/md_5/bungee/api/ProxyConfig.java
+@@ -168,6 +168,11 @@ public interface ProxyConfig
+      */
+     boolean isLogServerListPing();
+ 
++    /**
++     * Whether we log InitialHandler connections
++     */
++    boolean isLogInitialHandlerConnections();
++
+     /**
+      * The supported versions
+      */
+diff --git a/proxy/src/main/java/io/github/waterfallmc/waterfall/conf/WaterfallConfiguration.java b/proxy/src/main/java/io/github/waterfallmc/waterfall/conf/WaterfallConfiguration.java
+index 7dc3c2d9..6dc58ca2 100644
+--- a/proxy/src/main/java/io/github/waterfallmc/waterfall/conf/WaterfallConfiguration.java
++++ b/proxy/src/main/java/io/github/waterfallmc/waterfall/conf/WaterfallConfiguration.java
+@@ -26,6 +26,13 @@ public class WaterfallConfiguration extends Configuration {
+      */
+     private boolean logServerListPing = false;
+ 
++    /**
++     * Whether we log InitialHandler connections
++     * <p>
++     * Default is false (don't log)
++     */
++    private boolean logInitialHandlerConnections = false;
++
+     /**
+      * The supported versions displayed to the client
+      * <p>Default is a comma seperated list of supported versions. For example 1.8.x, 1.9.x, 1.10.x</p>
+@@ -60,6 +67,7 @@ public class WaterfallConfiguration extends Configuration {
+         config.load(false); // Load, but no permissions
+         metrics = config.getBoolean("metrics", metrics);
+         logServerListPing = config.getBoolean( "log_server_list_ping", logServerListPing );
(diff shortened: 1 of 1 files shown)


MarkTwoFive:
Have you already tried proposing this upstream? I'd say they could be interested in this too.

sgdc3:
@MarkL4YG I already seen that, they closed with the "Just make a log filter" excuse if i remember well.

jamierocks:
I want to clarify my position on making certain log messages configurable, seeing as this is the second PR in a short period regarding an issue.

After a great undertaking, Minecrell overhauled our logging situation - allowing plugins to use slf4j loggers, and using log4j2 under-the-hood. This overhaul also allows for server administrators to use custom log4j2 configurations.

Obviously if the issue is widespread, then the use of a log4j2 configuration is perhaps not the best option (in addition to the fact that not all server administrators are going to want to learn how to write log4j2 configurations) - this is where I fundamentally agree with every configurable logging PR, as they ease things for server administrators.

However, I do not want Waterfall to become plagued with special-casing code - that requires maintaining (not that this is a problem, I will happily maintain them - just only if they are justified). This is why I would much rather have seen PRs like this having gone through upstream first.

Should it be merged in Bungeecord - great, if not - I will consider merging the configuration option in Waterfall.

---

***EDIT**: Having spoke with Minecrell, we've decided that writing custom log4j2 configurations to filter the messages is not too great a solution - as they would have to be filtered by output, and not logger (as waterfall uses the same logger)*

*Perhaps the better solution here is to go through Waterfall and introduce local loggers as opposed to the current global logger situation (Bungeecord PR)*

BungeeCord-Patches/0046-Make-InitialHandler-connections-logging-optional-dis.patch
```diff
@@ -0,0 +1,77 @@
...
+@@ -26,6 +26,13 @@ public class WaterfallConfiguration extends Configuration {
+      */
+     private boolean logServerListPing = false;
+ 
++    /**
++     * Whether we log InitialHandler connections
++     * <p>
++     * Default is false (don't log)
++     */
++    private boolean logInitialHandlerConnections = false;
```
electronicboy: Default to true, don't disable logging messages which provide some level of information as to what is going on

jamierocks:
*Note this is by no means a final solution, and if a better more generic solution cannot be made that is not of too great a difficulty for administrators to do, this config option will remain in perpetuity.*

Having given this some thought - here's my solution:

- I'm going to write a pull request for Bungeecord, introducing the use of local Loggers (that would allow logger configurations to be written for this sort of thing) - regardless of whether it makes it way into Bungeecord, it will make its way into Waterfall
- Temporarily, I will merge this PR to provide an immediate solution to server administrators - deprecating it further down the line, when a full-time, generic solution is in place - proper notice will be given for this, and I will write this up in full on Aquifer.org later this year.

I'm no fan of removing configuration options, but given it will be deprecated before removal and there will be a more powerful alternative will be available.

*This PR will still receive the typical scrutiny that any PR would get.*

---

When I have made a more powerful solution available, I will begin writing resources for making use of it and make a notice marking the deprecation of the configuration option, and its removal date. I will provide example configurations that resolve the issue, to allow admins to easily transition to the new system and will potentially write some documentation for Waterfall (which is a long-time coming).

---

***tl;dr**: This config option will be available most of this year, until it gets removed (having had a more powerful replacement) having been deprecated for some time. The config option will likely be around for the majority of the year.*

**I CANNOT STRESS ENOUGH, THAT PROPER AND FULL NOTICE WILL BE GIVEN BEFORE THE OPTION IS REMOVED LATER IN THE YEAR** - *it will be deprecated long before it is removed*

BungeeCord-Patches/0046-Make-InitialHandler-connections-logging-optional-dis.patch
sgdc3: This log message tells you only one information:
Not under attack<- (YES) Can you read the console log? (NO) -> Under attack
😆

sgdc3: It makes sense only in debug environments

electronicboy: It's a message which can provide information on what is going on, which is the reason why the message is there in the first place. I don't agree with disabling information messages by default which can be useful should something be going on, just because you disagree with that notion.

sgdc3:
@jamierocks any news?

jamierocks:
Having given this some thought, there really isn't any alternate way of handling this in such a way as this as a config option. Therefore, I am quite happy to merge this once you've addressed electroniccat's review.

sgdc3:
Uh, nice i'll do that asap ;)

sgdc3:
@electronicboy ready for review

sgdc3:
Any news?

BungeeCord-Patches/0046-Make-InitialHandler-connections-logging-optional-dis.patch
```diff
@@ -0,0 +1,77 @@
...
++    private boolean logInitialHandlerConnections = true;
++
+     /**
+      * The supported versions displayed to the client
+      * <p>Default is a comma seperated list of supported versions. For example 1.8.x, 1.9.x, 1.10.x</p>
+@@ -60,6 +67,7 @@ public class WaterfallConfiguration extends Configuration {
+         config.load(false); // Load, but no permissions
+         metrics = config.getBoolean("metrics", metrics);
+         logServerListPing = config.getBoolean( "log_server_list_ping", logServerListPing );
++        logServerListPing = config.getBoolean( "log_initial_handler_connections", logInitialHandlerConnections );
```
jamierocks, line 46: Setting the wrong field

jamierocks (changes requested):

jamierocks:
Merged with https://github.com/WaterfallMC/Waterfall/commit/4c55f25c9ca27c71a4d4e7e6788d5d71c459dca8.

PR #222, opened 2018-04-24
Packet Performance API

Narimm:
This Patch adds a new api that allows a server owner to track packet statistics - this can allow server owners to track and monitor various packets and statistics relating to the communication between minecraft servers and Waterfall.

 - Recently it was used to diagnose and fine tune Spigot server settings relating to view distance and chunk loading to improve load times from players with high network transit times.
 - It does require a server owner to implement the api to retrieve the stats usually in the form of a plugin.
Eg: https://github.com/AddstarMC/BungeePandora/blob/dc9fb46ef2287261d713ad6370789fd24da658d8/src/main/java/au/com/addstar/bpandora/modules/Perfmon.java

This shows how we implement the tracker and store the stats in a mysql database but equally they could be reported to graphite or other statistics services

### DIFF
--- a/BungeeCord-Patches/0047-Added-packet-performance-stats-api.patch
+++ b/BungeeCord-Patches/0047-Added-packet-performance-stats-api.patch
@@ -1,0 +1,532 @@
+From 89c67b56dc5a5613216120aaa67ad38192bb60c5 Mon Sep 17 00:00:00 2001
+From: Narimm <benjicharlton@gmail.com>
+Date: Tue, 24 Apr 2018 15:38:39 +1000
+Subject: [PATCH] Added packet performance stats api
+
+* Allows you to see a breakdown of network traffic
+
+diff --git a/api/src/main/java/net/md_5/bungee/api/PerformanceStatistics.java b/api/src/main/java/net/md_5/bungee/api/PerformanceStatistics.java
+new file mode 100644
+index 00000000..295dbe52
+--- /dev/null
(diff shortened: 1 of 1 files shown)


mikroskeem:
If you are contributing this to Waterfall, you might want to use Waterfall package (io.github.waterfallmc) for new classes and API

Narimm:
To be honest we have been using this with Bungeecord for over 2 years...It was built to work with any bungeecord based fork.  I am disinclined to do the work to move it to waterfalls api, which limits me to making plugins that don't work on both.

Narimm:
That being said i completely understand if this means the dev team reject the PR

MarkTwoFive:
What does upstream say to your proposal?
I would like to comment that this kind of feature may be better suited for the main application rather than "just" a fork in my opinion.

Black-Hole:
Some kind of config option would be needed, so if disabled it won't even be added to the Netty pipeline.

The statistics are not thread safe.

There are also some formatting issue like using tabs instead of spaces.

mikroskeem:
> I am disinclined to do the work to move it to waterfalls api

Yeah you'd be better off to opening PR on upstream in that case.

Narimm:
I'll PR upstream and see what they say - I will be honest and say I didnt really bother I didnt figure the BC community to be really that interested in the  fine tuning.  This api allowed us to specifically determine what packets from which server and  I will close this down for now and await the response from upstream

PR #230, opened 2018-05-22
Upstream merge

Black-Hole:
Just a simple upstream merge without any conflicts to solve. It compiles.


electronicboy:
been meaning to touch this myself, not entirely fond of some of the upstream changes and wanted to see if they got any comments surrounding forge, as I know that forge support is disabled by default now, which seems like a bad default, especially as a fork which attempts to provide some sanity to forge...

jamierocks:
Sadly I don't have the time at the moment to make this merge (would require looking into the changes, etc). While I can agree with a toggle for Forge support, Waterfall will enable this by default.

We will need to go over all of our Forge-related patches to ensure that they work with the new toggles, in addition to looking into whatever other changes upstream have made.

If nobody does the merge in full by the weekend, I'll sort it out ;)

jamierocks:
I did say the weekend, so I'm only slightly late. Enjoy! https://github.com/WaterfallMC/Waterfall/commit/f539c023a30ff59a3c0890e8e1fc12c9af041711

PR #231, opened 2018-05-28
Update Maven Deploy URL

zachbr:
We recently moved some things around and as a result several URLs
changed.

Redirects are in place and most users will not need to make any changes.
We have already tested and confirmed that maven access works just fine
via the redirects.

However, maven will not allow you to deploy through the redirect so
projects using old deploy URLs will need to be updated.


jamierocks:
Awesome, thanks!

PR #236, opened 2018-06-17
Enable IPv6 Support

Schmorzel:
Minecraft supports IPv6

### DIFF
--- a/BungeeCord-Patches/0048-IPv6-enable.patch
+++ b/BungeeCord-Patches/0048-IPv6-enable.patch
@@ -1,17 +1,17 @@
 From: Tim Zimmermann <tim.zimmermann@server24-7.eu>
-Date: Sun, 17 May 2018 15:44:57 +0200
+Date: Sun, 17 May 2018 16:29:11 +0200
 Subject: [PATCH] Enable IPv6 Support
 
 diff --git a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
-index 07e46026..3a0f9550 100644
+index 77873239..63a28204 100644
 --- a/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
 +++ b/proxy/src/main/java/net/md_5/bungee/BungeeCord.java
-@@ -252,7 +252,7 @@ public class BungeeCord extends ProxyServer
+@@ -249,7 +249,7 @@ public class BungeeCord extends ProxyServer
      @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED_BAD_PRACTICE")
      public void start() throws Exception
      {
 -        System.setProperty( "java.net.preferIPv4Stack", "true" ); // Minecraft does not support IPv6
-+        //System.setProperty( "java.net.preferIPv4Stack", "true" ); Commented out this line to enable IPv6 support
++        //System.setProperty( "java.net.preferIPv4Stack", "true" ); // Waterfall - Enable IPv6 support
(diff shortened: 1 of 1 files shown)


Black-Hole:
Don't just remove the line, comment it out with a Waterfall marker comment at the end.

PhanaticD:
would prefer if this was configureable

Phoenix616:
Ipv6 should already be supported. This part only makes it prefer ipv4 if available.

EDIT: Wait, I might be wrong about that one. The [docs](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) say that it will not be possible to communicate with an IPv6 only host and that the default behaviour is already to prefer IPv4.

Black-Hole:
@PhanaticD You configure this using the network settings of your host and the listening host in the config.yml. Like in every other networking program.

Black-Hole:
@Schmorzel Have a look at other Waterfall commits. Every one uses "// Waterfall" comments on changed lines.

mikroskeem:
@Schmorzel

Your change should look like this:
`//System.setProperty( "java.net.preferIPv4Stack", "true" ); // Waterfall - Enable IPv6 support`

Also don't forget to squash your commits and then force push :smile:

jamierocks:
Yes, the patch should look a little different. Though I'm totally in favor of not making an issue out of it, I can clean up small style issues like this during the merge.

electronicboy (commented):
The only real question around something like this is, what impact, if any, would this have on existing servers? Bearing in mind that a fair chunk of people running servers are going to have a v6 on their LAN but not v6 externally, would this randomly break their setup?

Phoenix616:
Well according to the [docs](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) on the `java.net.preferIPv6Addresses` property it should prefer IPv4 over IPv6 if it's available to ensure backwards compatibility:

> When dealing with a host which has both IPv4 and IPv6 addresses, and if IPv6 is available on the operating system, the default behavior is to prefer using IPv4 addresses over IPv6 ones.

MarkTwoFive:
@electronicboy I want to add that more and more ppl are getting IPv6 connectivity on their private internet connection at least in Europe as we've already run out of available addresses. Therefore a setting which would allow users aware of the protocol to opt-in would be a viable option in my opinion.

electronicboy:
my only concern is around blindly breaking peoples setups, saying that, a little while ago I went back to work out why it was added and can't really see anything beyond it seemingly being added blindly, so long as this won't break existing setups which are somewhat common, it'll have my approval
Shamefully, I don't have a network setup which really allows me to test this without having to spend the time whipping up a few VMs

jamierocks:
One thing of note is that though Minecraft doesn't set that flag (at least not anymore, if it ever did - unless GH doesn't index repos over a certain size), Forge does. Though the Forge tweaker is about to disappear in 1.13, so maybe that will disappear with it?

I'm also not sure that the flag actually prevents ipv6 usage anyhow?

electronicboy:
http://www.minecraftforge.net/forum/topic/61760-no-ipv6-with-112-and-above/

Looks like it causes some issues potentially, due to the potential issues around not setting this, shove it behind a config option that sets it by default, and it'll be good by me 👍 

if it can be proven that not setting that in most common setups doesn't cause issues, we can consider removing the config option (and the setting of that entirely) in the future

chuushi:
Bungee removed the line today: https://github.com/SpigotMC/BungeeCord/commit/76dc32ee3259c9d52ce71467c926c055693f5e92

electronicboy:
Closing as this has been done upstream (and has been merged into waterfall)

PR #259, opened 2018-07-22
[1.13] Fix scoreboard team packet issues

mariaum:
Sometimes the server will send an empty component.
Not sure if i am following the right code style, or there is a better way of doing this.


electronicboy:
Cheers! looks fine to me, I've merged this in as-is for now

PR #261, opened 2018-07-22
Dispatch events using MethodHandles

jamierocks:
MethodHandles are faster than reflection, and improve exceptions (the exceptions are not through proxies).


Janmm14:
MethodHandles are considered slower. At least it was that case last time I checked the topic.

Also because the JVM generates direct access classes for normal reflection.

jamierocks:
Yeah, I've read a number of conflicting accounts of MethodHandles - will need to benchmark.

Leymooo:
Maybe use LambdaMetaFactory?
https://github.com/Leymooo/BungeeCord/commit/fa2c5738b7cd0d61913fe237b5c1084db02a57bf
https://github.com/Leymooo/BungeeCord/commit/d779d36eded6d689e002255f1c1516184d428c14

jamierocks:
RIP.

PR #269, opened 2018-07-27
1.13 scoreboard api support

Leymooo:
This patch will return a 1.13 scoreboard support for not updated plugins. Also will reduce a memory usage, because of legacy text is shorter, than json.


electronicboy:
Closing, while providing support for plugins which haven't updated would be great, this PR does this at the cost of potentially causing incompatibilities with the expectations that a plugin may have running on bungee.

as much as upstreams treatment (or lack thereof) around this API leaves much to be desired, we have to respect the behavior that their API provides unless we can prove that it's broken and a fix wouldn't impact properly behaving plugins. Even if we do agree that the handling of this is somewhat broken, plugins are going to adapt to this behavior, and I don't want waterfall to break around this due to their handling of it, even if their implementation may be wrong.

Leymooo:
This patch does not have any problems with plugins those was updated for upstream. 

This patch mainly for reduce a memory usage. Scoreboards are big problem in BungeeCord and they use the most memory.

PR #309, opened 2018-11-18
Start removing entity map rewriting

DoNotSpamPls:
This looked *very* easy to me, so there's a chance I'm a total moron and missed something. Let me know if that's the case.
Still, BungeeCord builds and hasn't crashed violently in my short testing, so that's something 😛


mikroskeem:
Isn't it easier just to comment out `EntityMap` usage instead of deleting that file completely?

Leymooo:
https://github.com/SpigotMC/BungeeCord/tree/alt-respawn
but you still need to reset BossBar's

DoNotSpamPls:
[line 202?](https://github.com/SpigotMC/BungeeCord/commit/f70d4e2295c1f4199e84521e48f635a30282a1c1#diff-e686ecd2fbb36b8bdb49ed488d8e783dR202)

Leymooo:
Need keep L226-231

Leymooo:
Also it looks like a respawn packet does not send if dimension is same. also need revert this change(keep only line 237).

DoNotSpamPls:
@Leymooo the thing is, I never removed the lines you're talking about, they are still there.

Leymooo:
You can not simple remove an entitymap rewrites. You need also sync a server entityID and a client entityID

BungeeCord-Patches/0046-Remove-entity-map-rewriting.patch
```diff
@@ -0,0 +1,61 @@
...
+
+diff --git a/proxy/src/main/java/net/md_5/bungee/UserConnection.java b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+index 9c872a1c..5e6e507f 100644
+--- a/proxy/src/main/java/net/md_5/bungee/UserConnection.java
++++ b/proxy/src/main/java/net/md_5/bungee/UserConnection.java
+@@ -134,8 +134,10 @@ public final class UserConnection implements ProxiedPlayer
+     /*========================================================================*/
+     @Getter
+     private String displayName;
+-    @Getter
```
mikroskeem, line 57: I also suggest using multiline comments
```
/* Waterfall start
@Getter
private EntityMap entityRewrite;
*/ // Waterfall end

Leymooo:
This should be a correct way of removing an entity rewriting(I hope). ~~https://github.com/Leymooo/BungeeCord/commit/515557f4320a0ba04d26c5a2691b52fa636812fb~~ ~~https://github.com/Leymooo/BungeeCord/commit/405e81e017fd7230620930cc035abcd030b3d8af~~ 
upd: https://github.com/Leymooo/BungeeCord/commit/49daeab851b6634dfdb7891caa944d329be7a8b1 it seems to work

Fejm:
There is no need to reset BossBar (alt-respawn branch is based on my idea)

Simply, send Login Packet on any server switch, this causes client to reset Scoreboards, BossBars, EntityID's etc.

You only need to implement BungeeCord API there. I mean, if you set for example scoreboard's via Bungee you should resend it on server switch. I just removed all of these APIs on my server because I do not use it.

DoNotSpamPls:
Okay, I'm seriously confused at this point. Could you give some example code I can look at in order to understand what you mean?

Fejm:
https://gist.github.com/Fejm/bfd76cd00e550e6797b89987cc0bfd7a

This is working server switching method. But there is no Scoreboard API (Bungee side) etc. I'm not using it so i just removed them.

Please note that this was copy - pasted from my custom fork. I just removed my "custom" things from there, so it may look bad ;)

DoNotSpamPls:
@Fejm I think I've implemented the code, although I haven't tested it yet (feeling a bit ill today). Check it out and tell me if it was implemented properly.

Leymooo:
@fejm are you sure that no need to reset bossbars? I don't checked it, but if i don't mistaken bossbars are stored in object those does not reset on login/respawn packets.

UPD. I checked it. It does not reset bossbars. I check it on 1.12.2

Leymooo:
@DoNotSpamPls i think you dont understand why bungeecord ~~needs~~ uses entity rewrite.

When player joins, server sends login(join game) packet, this packet contains player enitity id. When we switch server, server also sends login packet, but bungeecord does not forward it to player. This means that server and client have different player entity id. BungeeCord uses entity rewrite to sync entity ids. (it just process all packets those contains entity id and replace id with correct one)

**But** minecraft allows to send login packet many times. this means that we can simple send login packet again to sync player and server entity id and dont use entity rewrite.  I gave to you a link above to **alt-respawn** branch, where this implemented. Also i gave a link to my commit where should be a correct implementation.

PR #345, opened 2019-02-01
Improved TabCompleteResponseEvent

Joo200:
- Give access to Brigadier's Suggestions object.
- Access to the list of suggestions through a self written list class

### Reasons for the change:
The TabCompleteResponseEvent from BungeeCord has a new behavior with Minecraft's Brigadier:
1. Load all TabCompletions from spigot in the Suggestions object.
2. Move the text from all TabCompletions in a ArrayList.
3. Call the TabCompleteResponseEvent
4. If the event changed single TabCompletions: Create a new Suggestions object from the ArrayList. Drop all existing Tooltips.

There will probably be no plugin that offers tooltips. But it will only be a matter of time before problems arise with this handling.

I tested it with Spigot 1.13.2 and Spigot 1.12.2

### Questions:
- Should I add some test cases?
- This might be an interesting idea for TabCompleteEvent.
- Add a config variable to enable this handling?

### DIFF
--- a/BungeeCord-Patches/0048-Improved-TabCompleteResponseEvent.patch
+++ b/BungeeCord-Patches/0048-Improved-TabCompleteResponseEvent.patch
@@ -1,0 +1,413 @@
+From a688607cad823893ee8498c3a96408702efdee2a Mon Sep 17 00:00:00 2001
+From: JOO200 <git@johanneszangl.de>
+Date: Fri, 1 Feb 2019 00:41:55 +0100
+Subject: [PATCH] Improved TabCompleteResponseEvent: * Give access to
+ Brigadier's Suggestions object. * Access to the list of suggestions through a
+ self written list class
+
+
+diff --git a/api/src/main/java/io/github/waterfallmc/waterfall/utils/SuggestionList.java b/api/src/main/java/io/github/waterfallmc/waterfall/utils/SuggestionList.java
+new file mode 100644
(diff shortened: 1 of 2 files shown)


electronicboy:
Sorry for the delay, kept starting to reply on this and I just never seem to finish

Two main points, 

The first being that we shouldn't be exposing brigadier objects (brigadiers only accessible in the API due to it being used in the protocol, ideally brigadier shouldn't be visible here), they're not ours and we cannot guarantee any form of consistency, as such they are not suitable for API here

currently ill so don't really want to comment on every issue in SuggestionList, but last I looked there was one or two methods which looked wrong, e.g. `toArray(T[] a)` does not conform to behavior expected from List

Joo200:
Okay, instead of showing brigadier there should be an own implementation of suggestions.
I understand the point of consistency.

Do you think the implementation of a "forwarding" list is acceptable?
I don't want to invest time in this solution only to get a denied PR ;)

compieter-gh:
I have seen an issue where if a server send the tabcompletelist too bungeecord it doesn't reset on server switch which should happen and request a new tabcompletelist from the joined server am i correct?

Joo200:
ComPieter1, that doesn't have anything to do with that pull request.

I forgot totally about that PR. As there are some better ideas (Paper's brigadier implementation) I'll close that PR.

PR #359, opened 2019-03-11
Things to the API patch

MrIvanPlays:
This patch adds some things to the API. It adds feature to schedule tasks with ticks, and adds some things to the player interface, also removes 2 non reason deprecations.

### DIFF
--- a/BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
+++ b/BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
@@ -1,0 +1,264 @@
+From fa435a332e99ec02e25ffb65d90c78aa10bbd172 Mon Sep 17 00:00:00 2001
+From: MrIvanPlays <pekov.ivan@abv.bg>
+Date: Mon, 11 Mar 2019 16:00:17 +0200
+Subject: [PATCH] Add some things to the API
+
+
+diff --git a/api/src/main/java/net/md_5/bungee/api/Callback.java b/api/src/main/java/net/md_5/bungee/api/Callback.java
+index 0cccc175..f383279b 100644
+--- a/api/src/main/java/net/md_5/bungee/api/Callback.java
++++ b/api/src/main/java/net/md_5/bungee/api/Callback.java
+@@ -6,6 +6,7 @@ package net.md_5.bungee.api;
+  *
+  * @param <V> the type of result
+  */
++@FunctionalInterface // Waterfall
+ public interface Callback<V>
+ {
+ 
+diff --git a/api/src/main/java/net/md_5/bungee/api/CommandSender.java b/api/src/main/java/net/md_5/bungee/api/CommandSender.java
+index a35b3fd0..f83fc023 100644
+--- a/api/src/main/java/net/md_5/bungee/api/CommandSender.java
++++ b/api/src/main/java/net/md_5/bungee/api/CommandSender.java
+@@ -16,19 +16,19 @@ public interface CommandSender
+ 
+     /**
+      * Send a message to this sender.
++     * Waterfall - Deprecation removal. No reason deprecating
+      *
+      * @param message the message to send
+      */
+-    @Deprecated
+     public void sendMessage(String message);
+ 
+     /**
+      * Send several messages to this sender. Each message will be sent
+      * separately.
++     * Waterfall - Deprecation removal. No reason deprecating
+      *
+      * @param messages the messages to send
+      */
+-    @Deprecated
+     public void sendMessages(String... messages);
+ 
+     /**
+diff --git a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+index f7459860..9946c310 100644
+--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
++++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+@@ -69,7 +69,9 @@ public interface ProxiedPlayer extends Connection, CommandSender
+      *
+      * @param position the screen position
+      * @param message the message to send
++     * @deprecated This is unlikely the API you want to use. See {@link #sendActionBar(String)} or {@link #sendActionBar(BaseComponent)}, {@link #sendActionBar(BaseComponent...)} for a more proper ActionBar API.
+      */
++    @Deprecated // Waterfall
+     public void sendMessage(ChatMessageType position, BaseComponent... message);
+ 
+     /**
+@@ -77,7 +79,9 @@ public interface ProxiedPlayer extends Connection, CommandSender
+      *
+      * @param position the screen position
+      * @param message the message to send
++     * @deprecated This is unlikely the API you want to use. See {@link #sendActionBar(String)} or {@link #sendActionBar(BaseComponent)}, {@link #sendActionBar(BaseComponent...)} for a more proper ActionBar API.
+      */
++    @Deprecated // Waterfall
+     public void sendMessage(ChatMessageType position, BaseComponent message);
+ 
+     /**
+@@ -372,4 +376,37 @@ public interface ProxiedPlayer extends Connection, CommandSender
+      * @return this player's {@link Scoreboard}
+      */
+     Scoreboard getScoreboard();
++
++    // Waterfall start
++
++    /**
++     * Sends ActionBar message to the client.
++     *
++     * @param message the message you wish to send
++     */
++    public void sendActionBar(String message);
++
++    /**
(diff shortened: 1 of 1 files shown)


BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
+@@ -16,19 +16,19 @@ public interface CommandSender
+ 
+     /**
+      * Send a message to this sender.
++     * Waterfall - Deprecation removal. No reason deprecating
+      *
+      * @param message the message to send
+      */
+-    @Deprecated
+     public void sendMessage(String message);
```
hugmanrique, line 32: This method is deprecated for valid reasons. Legacy texts are converted to text components, but you should always use the latter.

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
+ 
+     /**
+      * Send several messages to this sender. Each message will be sent
+      * separately.
++     * Waterfall - Deprecation removal. No reason deprecating
+      *
+      * @param messages the messages to send
+      */
+-    @Deprecated
```
hugmanrique, line 41: Same thing here

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
+--- a/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
++++ b/api/src/main/java/net/md_5/bungee/api/connection/ProxiedPlayer.java
+@@ -69,7 +69,9 @@ public interface ProxiedPlayer extends Connection, CommandSender
+      *
+      * @param position the screen position
+      * @param message the message to send
++     * @deprecated This is unlikely the API you want to use. See {@link #sendActionBar(String)} or {@link #sendActionBar(BaseComponent)}, {@link #sendActionBar(BaseComponent...)} for a more proper ActionBar API.
+      */
++    @Deprecated // Waterfall
+     public void sendMessage(ChatMessageType position, BaseComponent... message);
```
hugmanrique, line 56: This method is perfectly valid, why would we deprecate it?

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
+ 
+     /**
+@@ -77,7 +79,9 @@ public interface ProxiedPlayer extends Connection, CommandSender
+      *
+      * @param position the screen position
+      * @param message the message to send
++     * @deprecated This is unlikely the API you want to use. See {@link #sendActionBar(String)} or {@link #sendActionBar(BaseComponent)}, {@link #sendActionBar(BaseComponent...)} for a more proper ActionBar API.
+      */
++    @Deprecated // Waterfall
+     public void sendMessage(ChatMessageType position, BaseComponent message);
```
hugmanrique, line 66: Same thing as above.

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
+index 81bd18f0..44ff7cd9 100644
+--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
++++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+@@ -181,10 +181,14 @@ public class PluginManager
+             {
+                 if ( proxy.getConfig().isLogCommands() )
+                 {
+-                    proxy.getLogger().log( Level.INFO, "{0} executed command: /{1}", new Object[]
++                    // Waterfall - check if the command sender is not the console
++                    if ( !sender.getName().equalsIgnoreCase( proxy.getConsole().getName() ) )
```
hugmanrique, line 135: Logging console commands might be useful for web terminals/screen/tmux... where multiple people might be watching the same console output

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
++
++    /**
++     * Schedules a task to be executed asynchronously after the specified delay
++     * is up.
++     *
++     * @param owner the plugin owning this task
++     * @param task the task to run
++     * @param delay the delay before this task will be executed, specified in ticks
++     * @return the scheduled task
++     */
```
hugmanrique, line 165: Bungee (and by extension, Waterfall) doesn't have a concept of ticks. This is only used in NMS based servers. You can pass 50 millis if you want to run a task every tick.

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
++    @Getter
++    private Title currentTitle;
+     // Waterfall end
+     /*========================================================================*/
+     @Getter
+@@ -709,6 +711,7 @@ public final class UserConnection implements ProxiedPlayer
+     @Override
+     public void sendTitle(Title title)
+     {
++        this.currentTitle = title; // Waterfall
```
hugmanrique, line 204: I don't see any current title expiration code

hugmanrique (changes requested):
I like the additions of sendActionBar and sendTitle, but the other stuff would require too much maintenance + it should be in another separate PR (and patch)

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
Janmm14: And generally useful when checking the log file?

Janmm14 (commented):
Heck, we have the fancy sendMessage with chat position. I don't see these sendActionBar function a useful abbreviation. It might even add further confusion in combination with semdMessage with action bar position.

I'd say waterfall should not encourage alternatives to working api in bungee.

And btw. there are 3 message types. When you deprecate the function with the enum and don't add functions with chat and system type difference, its not good.

Janmm14:
And "Add things to the API" as description is also not ok. What you did here belongs into multiple commits/patches with more descriptive naming. (Apart from most changes I and the previous reviewer think are not good)

hugmanrique:
Yeah, on a second thought, adding an extra method every time a new chat position is created is not scalable.

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
```diff
@@ -0,0 +1,264 @@
...
+     Scoreboard getScoreboard();
++
++    // Waterfall start
++
++    /**
++     * Sends ActionBar message to the client.
++     *
++     * @param message the message you wish to send
++     */
++    public void sendActionBar(String message);
```
kashike, line 81: Add `@Deprecated` to this method.

kashike (changes requested):

MrIvanPlays:
That's why i didn't wanted to PR, but everyone said 'just PR to waterfall'. I KNEW that it will be rejected...

aikar:
You need to be able to take feedback. The quality of your submission was low, and people have provided feedback on what was wrong so that you can learn to improve the quality of your contributions.

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
mdcfe: In addition, you should specify either in Javadocs or the method name (or both) that this method accepts ticks. It isn't obvious what unit the delay is in unless you read the implementation code.

BungeeCord-Patches/0049-Add-some-things-to-the-API.patch
MrIvanPlays: Then why md_5 doesn't specify it in javadocs as '@ deprecated' ?!

mdcfe: You can already see it's deprecated from the annotation on the method; he probably felt it wasn't necessary to duplicate that line.

aikar: Honestly, Deprecating string methods is stupid, as it serves a purpose to send a simple message.

One could argue the legacy codes support of this message are 'deprecated' and that the method isn't, but it's also possible to permanently support that system on the API, which I'm supportive of, as components are stupid for simple messaging. 

I'm in favor of permanently keeping legacy codes around for simple messaging, and if mojang adds more advanced capabilities than current with components, then the legacy system just wouldn't expose it and people can move to comps then.

But even a sendMessage(ChatColor color, String msg); that doesn't use legacy codes would be useful to declare simple messages, but I see no harm in keeping legacy codes on sendMessage.

If md5 removed it from Bukkit API, I would add it back. It costs nothing to maintain.

PR #362, opened 2019-03-18
Apply delombok to sources, generate javadocs from delombok sources

Byteflux:
Very simple change to delombok sources and create javadocs from the generated sources.


aikar (approved):
Looks good to me. Solves a major complaint people have about lombok :P

electronicboy (approved):

PR #373, opened 2019-04-17
Speed up packet construction

mikroskeem:
According to @electronicboy's benchmark, this is 6x faster than plain reflection.


Aurelien30000:
You should implement that, i think it's a good idea.

Janmm14:
The exception handling is wrong in this edit, it changes behaviour without reason.

1. Instead of retuening null if the constructor supplier is absent, you have to throw a BadPacketException.
2. The try-catch needs to be removed.

Janmm14 (changes requested):
The exception handling is wrong in this edit, it changes behaviour without reason.

1. Instead of returning null if the constructor supplier is absent, you have to throw a BadPacketException.
2. The try-catch needs to be removed (not 100% sure).

mikroskeem:
>Instead of returning null if the constructor supplier is absent, you have to throw a BadPacketException.

[Wouldn't throwing BadPacketException change the behaviour](https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/Protocol.java#L410)? BungeeCord does not throw anything if constructor is not present, so I shouldn't throw either.

BungeeCord-Patches/0054-Speed-up-packet-construction.patch
```diff
@@ -0,0 +1,337 @@
...
+-            } catch ( ReflectiveOperationException ex )
++                return ( constructor == null ) ? null : constructor.get(); // Waterfall - speed up packet construction
++            } catch ( Exception ex ) // Waterfall - speed up packet construction
+             {
+                 throw new BadPacketException( "Could not construct packet with id " + id, ex );
+             }
+         }
+ 
+-        private void registerPacket(Class<? extends DefinedPacket> packetClass, ProtocolMapping... mappings)
++        public <P extends DefinedPacket> void registerPacket(Class<P> packetClass, java.util.function.Supplier<P> constructor, ProtocolMapping... mappings) // Waterfall - speed up packet construction
```
mikroskeem: private -> public is a leftover from my fork, I didn't intend to change access modifiers.

BungeeCord-Patches/0054-Speed-up-packet-construction.patch
```diff
@@ -0,0 +1,337 @@
...
+                     data.packetConstructors[mapping.packetID] = constructor;
+                 }
+-            } catch ( NoSuchMethodException ex )
++            } catch ( Exception ex ) // Waterfall - speed up packet construction
+             {
+                 throw new BadPacketException( "No NoArgsConstructor for packet class " + packetClass );
+             }
+         }
++        // Waterfall start - speed up packet construction (backwards compat)
++        public <P extends DefinedPacket> void registerPacket(Class<P> packetClass, ProtocolMapping... mappings) {
```
mikroskeem: Ditto (should not be public)

Janmm14:
@mikroskeem Sorry, I misread the old code, ofc a missing constructor should not throw since bungee doesn't handle all packet types.

mikroskeem:
Bump

colinrgodsey (approved):
this all LGTM. might end up pulling it into Waterdog if it doesnt end up getting merged here

electronicboy (approved):

BungeeCord-Patches/0054-Speed-up-packet-construction.patch
```diff
@@ -0,0 +1,347 @@
...
+                 Constructor<? extends DefinedPacket> constructor = packetClass.getDeclaredConstructor();
++                */ // Waterfall end
+ 
+                 int mappingIndex = 0;
+                 ProtocolMapping mapping = mappings[mappingIndex];
+@@ -479,11 +519,32 @@ public enum Protocol
+                     data.packetMap.put( packetClass, mapping.packetID );
+                     data.packetConstructors[mapping.packetID] = constructor;
+                 }
++            // Waterfall start - speed up packet construction
```
electronicboy, line 316: Removed the try/catch, this woulda incorrectly wrapped exceptions here

mikroskeem: Whoops, overlooked that detail :eyes: but thanks

PR #395, opened 2019-06-20
Properly check that the channel is closed

xxDark:


Black-Hole:
What issue are you trying to fix?
How do you confirm that it's fixed?

xxDark:
> 
> 
> What issue are you trying to fix?
> How do you confirm that it's fixed?

Bungeecord sometimes continues to write to the channel, even if it closed.
attachment

xxDark:
You can see it by yourself by logging all the exceptions in `HandlerBoss#exceptionCaught`

colinrgodsey:
I have noticed this before. Need to fully understand when TCP channels go inActive vs closing... I believe the 'active' state changes immediately, then 'close' doesn't fulfill until the underlying socket itself is closed (which may have to wait on a selector channel).

MrIvanPlays:
`// Waterfall` comment is missing when you do that change.

electronicboy:
This has already made it upstream, oh so long ago...
#LongLiveTheCat

PR #422, opened 2019-08-25
Make cmd-server module messages more configurable

ghost:
This is but a small change to the existing module (cmd-server) that would make a server list message more configurable. There are a lot of servers (including my own) that use this module to provide an easy way to change servers. While the module itself is really small and fast, it doesn't offer a full way to change the server list message. With this PR accepted people will be able to fully customize the message (including message after the end of the list and possibility to have different colors on comma or servers after the first one) without having to edit the source code.

There are people out there that want to make their in-game messages as perfect as possible, and this PR will help them achieve this.

The change itself doesn't impact the performance at all.


HexedHero (approved):

Janmm14:
This would likely get approved by bungeecord itself.

ghost:
> This would likely get approved by bungeecord itself.

I didn't think about that. I'm gonna close this PR for now and reopen it if BungeeCord doesn't accept a similar PR for some reason.

PR #425, opened 2019-09-14
Add gplugins command

MrIvanPlays:


sgdc3:
Nice, but i think modules are being deprecated...

electronicboy:
I do not believe that it's in the interest of the project to accept PRs from members who have been removed from the community

Issue #413, opened 2019-08-03
lost connection when chat only fullwidth space


tomo1560:
lost connection when chat only full-width space ("　") 
use version：waterfall#290
console log：https://pastebin.com/JZ94cpj1
image：
![2019-08-04_02-20-55](https://user-images.githubusercontent.com/12862158/62415030-971dbb80-b65e-11e9-8b45-ca477c02fbf1.gif)

Black-Hole:
This seems to be designed this way:
https://github.com/PaperMC/Waterfall/blob/master/BungeeCord-Patches/0024-Validate-that-chat-messages-are-non-blank.patch

electronicboy:
This is intentional, not 100% on why it was done, but there is a good chance that it was treated differently in older client versions than it is now and thus made sense to prevent other issues

Janmm14:
Should be changed probably to only check is empty in higher client versions.

PR #465, opened 2020-03-14
Do not throw IllegalArgumentException when empty chat has sent

rz7d:
Replaces a Preconditions API call with the simple if/return statement.
This Pull Request fixes Issue #413 

### Before
Throws IllegalArgumentException if received an empty chat.

### After
The empty chat is **ignored quietly**.

### DIFF
--- a/BungeeCord-Patches/0024-Validate-that-chat-messages-are-non-blank.patch
+++ b/BungeeCord-Patches/0024-Validate-that-chat-messages-are-non-blank.patch
@@ -48,7 +48,7 @@ index 4bb61f63..9287d168 100644
      {
          int maxLength = ( con.getPendingConnection().getVersion() >= ProtocolConstants.MINECRAFT_1_11 ) ? 256 : 100;
          Preconditions.checkArgument( chat.getMessage().length() <= maxLength, "Chat message too long" ); // Mojang limit, check on updates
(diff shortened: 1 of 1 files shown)


Artuto:
I dont think a call to a method should be ignored silently

rz7d:
> I dont think a call to a method should be ignored silently

@Artuto
Yes, I think so too.

I followed the commit history. But the original patch is created by [Initial Commit](https://github.com/PaperMC/Waterfall/blob/3809083007dd97e3a947ed0b20cd6523c420f4db/BungeeCord-Patches/0034-Validate-that-chat-messages-are-non-blank.patch).
I need more help or descriptions for this patch.
Why created the original patch?

I'll add the patch that logs empty messages.

Spigot 1.15.2 logs empty messages as warnings, but does not throw:
image

Anyway, throwing the exceptions is not suitable in this case.

electronicboy:
I was iffy on doing this myself in the past as I wasn't sure of the /why/ of this in the first place
@astei Are you able to give any insights on this?

Janmm14:
Not a proper solution to the problem.

ghost:
Still better than wasting resources on a exception that someone can exploit by just sending tons of null packet chats

Outfluencer:
> 
> 
> Still better than wasting resources on a exception that someone can exploit by just sending tons of null packet chats

How should this be exploitable? on the first empty chat packet you send, your connection will be closed instantly

PR #488, opened 2020-04-22
Add exception reporting event

theminecoder:
Since bungeecord has gone and made the plugin manager a final class (which is a good thing in some regards) there is no way to wrap around it to handle exception catching for commands and events.

This patch copies over the exception event (and corresponding exception classes) that exists in paper today and implements it in the places around the proxy codebase that should fire the event.

Couple of notes:
- Bungee's EventBus registration doesn't ever store a reference to the plugin that registered the listener so that exception wrapper doesn't have a plugin associated with it. (not quite something I was happy to play with on my own)
- `ProxyInternalException` & `ProxyPluginMessageException` aren't currently used anywhere but were left in the pr in case something needs to use them that I am unaware of.

### DIFF
--- a/BungeeCord-Patches/0055-Add-exception-reporting-event.patch
+++ b/BungeeCord-Patches/0055-Add-exception-reporting-event.patch
@@ -1,0 +1,699 @@
+From 9f488c256a9738d00aca1417cff6765090c877a4 Mon Sep 17 00:00:00 2001
+From: theminecoder <theminecoder.dev@gmail.com>
+Date: Wed, 22 Apr 2020 14:00:44 +1000
+Subject: [PATCH] Add exception reporting event
+
+
+diff --git a/api/src/main/java/io/github/waterfallmc/waterfall/event/ProxyExceptionEvent.java b/api/src/main/java/io/github/waterfallmc/waterfall/event/ProxyExceptionEvent.java
+new file mode 100644
+index 00000000..ee6cb5dd
+--- /dev/null
++++ b/api/src/main/java/io/github/waterfallmc/waterfall/event/ProxyExceptionEvent.java
+@@ -0,0 +1,27 @@
++package io.github.waterfallmc.waterfall.event;
++
++import com.google.common.base.Preconditions;
++import io.github.waterfallmc.waterfall.exception.ProxyException;
++import net.md_5.bungee.api.plugin.Event;
++
++/**
++ * Called whenever an exception is thrown in a recoverable section of the server.
++ */
++public class ProxyExceptionEvent extends Event {
++
++    private ProxyException exception;
++
++    public ProxyExceptionEvent(ProxyException exception) {
++        this.exception = Preconditions.checkNotNull(exception, "exception");
++    }
++
++    /**
++     * Gets the wrapped exception that was thrown.
++     *
++     * @return Exception thrown
++     */
++    public ProxyException getException() {
++        return exception;
++    }
++
++}
+diff --git a/api/src/main/java/io/github/waterfallmc/waterfall/exception/ProxyCommandException.java b/api/src/main/java/io/github/waterfallmc/waterfall/exception/ProxyCommandException.java
+new file mode 100644
+index 00000000..644a4cd6
+--- /dev/null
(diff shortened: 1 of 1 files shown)


Janmm14:
We need something that prevents stackoverflow: if an error happens during event handling of any event and if there is a faulty plugin always throwing in ProxyExceptionEvent this creates unneccessary long console output and cpu usage.

This could be achieved by having a separate, private callExceptionEvent method which requires a ProxyExceptionEvent and which gives the eventbus an empty exception handler and which is called instead of callEvent for ProxyExceptionEvents.

theminecoder:
> We need something that prevents stackoverflow: if an error happens during event handling of any event and if there is a faulty plugin always throwing in ProxyExceptionEvent this creates unneccessary long console output and cpu usage.
> 
> 
> 
> This could be achieved by having a separate, private callExceptionEvent method which requires a ProxyExceptionEvent and which gives the eventbus an empty exception handler and which is called instead of callEvent for ProxyExceptionEvents.

There is already a bailout check in the error handler that will return if we are handling a execution event, preventing the stackoverflow. 

https://github.com/PaperMC/Waterfall/pull/488/files#diff-10c042e7d4737de4aeaf80dc1fb8c89fR541

MrIvanPlays:
I was the one who made the PluginManager final. This was intentional to stop people from modifying it at runtime. Also I don't find a usecase about your pr; exceptions occurring in event handling or command handling should be handled there. A direct example is enforcing an specific argument type for command:

```java
public void execute(CommandSender sender, String[] args) {
   // ...
  try {
      Integer.parseInt(args[0]);
  } catch (NumberFormatException e) {
      sender.sendMessage(new ComponentBuilder("First argument should be number!").setColor(ChatColor.RED).build());
      return;
  }

  int firstArgument = Integer.parseInt(args[0]);
  // ...
}
```

For other things such as enabling/disabling/loading plugins on-the-fly it would be better to have a PR here or to upstream adding such feature to the api and not using hacks.

electronicboy (changes requested):
Looks fine, only real thing is the waterfall comments, they should annotate the set of changes, I'd suggest glancing over the paper contribution guide if you haven't

theminecoder:
Think I got everything with comments now. Let me know if I missed anything

theminecoder:
@MrIvanPlays I understand why you made it final. Honestly I think it's a good thing in the long run. However there is a use case to having exceptions being available like this - which is actually what this pr is about.

I make and use a plugin that grabs all of the exceptions from proxies and backend servers and forwards them to sentry so I can proactively keep an eye on issues without having to tail x amount of logs for all the servers that I run. Has been a lifesaver numerous amounts of time.

Until this pr is merged the only way to do this on bungee and waterfall is to proxy the plugin manager and even that doesn't get you access to scheduled tasks, so having the events allows me to monitor everything without hacky access to the server internals.

MrIvanPlays:
> @MrIvanPlays I understand why you made it final. Honestly I think it's a good thing in the long run. However there is a use case to having exceptions being available like this - which is actually what this pr is about.
> 
> I make and use a plugin that grabs all of the exceptions from proxies and backend servers and forwards them to sentry so I can proactively keep an eye on issues without having to tail x amount of logs for all the servers that I run. Has been a lifesaver numerous amounts of time.
> 
> Until this pr is merged the only way to do this on bungee and waterfall is to proxy the plugin manager and even that doesn't get you access to scheduled tasks, so having the events allows me to monitor everything without hacky access to the server internals.

I don't really understand what you're trying to achieve here. You want to collect all the exceptions in one place, no? Well, if thats the case then good idea for adding these events, however I still think that occurring exceptions should be handled in the source (command, event handling) and not to be suppressed in a event.

lucko:
> however I still think that occurring exceptions should be handled in the source (command, event handling) and not to be suppressed in a event.

Nobody is making the argument that they shouldn't?

Exception **handling** clearly isn't the use-case of the PR. Read the title: "exception **reporting**".

The exceptions are not being suppressed, they're just being exposed in an event as well as in a call to the proxy logger.

electronicboy (approved):

PR #501, opened 2020-05-08
Save memory on scoreboard teams

mikroskeem:
By using enums of course

### DIFF
--- a/BungeeCord-Patches/0056-Save-memory-on-scoreboard-teams.patch
+++ b/BungeeCord-Patches/0056-Save-memory-on-scoreboard-teams.patch
@@ -1,0 +1,117 @@
+From 3c091217e7879943cb5d7e3458fec93e9d74f29a Mon Sep 17 00:00:00 2001
+From: Mark Vainomaa <mikroskeem@mikroskeem.eu>
+Date: Fri, 8 May 2020 21:39:39 +0300
+Subject: [PATCH] Save memory on scoreboard teams
+
+
+diff --git a/api/src/main/java/net/md_5/bungee/api/score/Team.java b/api/src/main/java/net/md_5/bungee/api/score/Team.java
+index 39f81fd1..9b762778 100644
+--- a/api/src/main/java/net/md_5/bungee/api/score/Team.java
++++ b/api/src/main/java/net/md_5/bungee/api/score/Team.java
+@@ -17,8 +17,8 @@ public class Team
+     private String prefix;
+     private String suffix;
+     private byte friendlyFire;
+-    private String nameTagVisibility;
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

BungeeCord-Patches/0056-Save-memory-on-scoreboard-teams.patch
```diff
@@ -0,0 +1,107 @@
...
++    // Waterfall start - save memory on scoreboard teams
++    @Deprecated
++    public String getNameTagVisibility() {
++        return nameTagVisibility.getValue();
++    }
++
++    @Deprecated
++    public void setNameTagVisibility(String value) {
++        NametagVisibility enumValue;
++        if (value == null) {
```
Janmm14, line 37: I don't see a reason to add any safeties with an alternate, possibly wrong state here. Suggesting to leave out the null check and throw an exception if VALUES.get() returns null (same for collisionrule).

BungeeCord-Patches/0056-Save-memory-on-scoreboard-teams.patch
```diff
@@ -0,0 +1,107 @@
...
++        ALWAYS("always"),
++        PUSH_OTHER_TEAMS("pushOtherTeams"),
++        PUSH_OWN_TEAM("pushOwnTeam"),
++        NEVER("never"),
++        ;
++
++        @lombok.Getter
++        private final String value;
++
++        public static final java.util.Map<String, CollisionRule> VALUES = new java.util.HashMap<>();
```
Janmm14: Suggesting to use ImmutableMap or sth. similar (same for collisionrule)

BungeeCord-Patches/0056-Save-memory-on-scoreboard-teams.patch
mikroskeem: Right, will do so.

Janmm14: I edited my comment.

BungeeCord-Patches/0056-Save-memory-on-scoreboard-teams.patch
aikar: Eh i don't think the complexity in the change is worth that. immutable isnt gonna offer any real performance benefit.

Janmm14: Yeah, if there are 10 fields more or less in that array it won't make any difference in performance, I just thought it'd look better.

mikroskeem:
Made requested changes.

Issue #84, opened 2016-08-26
Issue with banned player


caoli5288:
Reproduction:
1. Ban a player
2. Player try join after banned
3. Un-ban it.
4. Player try join and got `You a already connected to this proxy`

Then the player can't join anymore. No issue with origin BungeeCord.

kamcio96:
plugins, Waterfall version?

jamierocks:
Probably related to https://github.com/WaterfallMC/Waterfall/issues/69

caoli5288:
@kamcio96 Latest Waterfall. Would happen with no plugin. Minecraft server version spigot-1.8.8

jamierocks:
It's possible its an issue with https://github.com/WaterfallMC/Waterfall/blob/master/BungeeCord-Patches/0027-Improve-connection-closing-fixing-the-kick-delay.patch

caoli5288:
@jamierocks I revert UserConnection.java. No effect. :-(

jamierocks:
Hmm, there are a lot of logic changes all over the place time to find where the cock up is :s

caoli5288:
@jamierocks I'll test revert your commit one by one...

caoli5288:
It's Event4j's issue. I hava a plugin that query which target player disconnect from but while it banned `null` returned and an NPE thrown. I add debug info into `BungeeCord.removeConnection` and see it called in BungeeCord but no called in Waterfall.
Seem like Event4J not catch all exception and break logic. @jamierocks @Techcable

jamierocks:
Interesting @Techcable are you able to fix anytime soon?

jamierocks:
I've removed the Event4J patch temporarily - https://github.com/WaterfallMC/Waterfall/commit/80bed7979d707563283635998cc326f6aa0ad5b0

jamierocks:
This issue as it is is resolved, Event4J will be reapplied when that is fixed. Resolved.

Issue #70, opened 2016-08-10
fake players after bot attack.


Leymooo:
video: https://youtu.be/IUmf2xY-hdE
screens:
![928eaea21a5e4c11ae1665747ef2cafe](https://cloud.githubusercontent.com/assets/8667294/17554613/6261fa7c-5f14-11e6-90c6-96e75754acc1.png)
![](https://i.gyazo.com/a53a7868d09af4fbccbaa2dda29059ae.png)
antibot plugin src: 
https://github.com/Dimatert9/AntiBot

mgl:
Same problem here: #69

jamierocks:
As with #69 I'm going to need the list of plugins you have installed, along with the build of Waterfall you're running and (if you've had any) stacktraces.

Leymooo:
only one plugin. AntiBot. Source in main post.
Latest build - https://github.com/Dimatert9/Waterfall-1

phase:
@Dimatert9 Open sourcing a list of your IPs wasn't the best idea.

Leymooo:
This is public anti bot for Russian users.

Leymooo:
Caused by connection throttle. https://youtu.be/u0SCVTCm3gc

mgl:
@Dimatert9 is right, setting connection throttle to 0 resolves the problem.

Janmm14:
As far as I know, Waterfall does not alter the connection-throttle. Does it happen with enabled connection throttle on normal bungeecord?

mgl:
It doesn't happen on normal bungee with exact same settings.

Leymooo:
Waterfall dont close connection and dont throw error if in PlayerDisconnectEvent was error.
Bungeecord log:
http://pastebin.com/ribLQ8Zs
WaterFall log.
18:10:21 [INFO] [/127.0.0.1:56541|Leymooooooooooo] <-> InitialHandler has connected
18:10:22 [INFO] [Leymooooooooooo] disconnected with: §cCould not connect to default or fallback server, please try again later: java.net.ConnectException
18:10:34 [INFO] [/127.0.0.1:56621] <-> InitialHandler has connected

there is no error, and UpstreamBridge has disconnected.

//plugin issue.

caoli5288:
Same as #84

PR #502, opened 2020-05-08
Use ASM to invoke event handlers

mikroskeem:


BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
```diff
@@ -0,0 +1,258 @@
...
++        {
++            MethodVisitor mv = cw.visitMethod(Opcodes.ACC_PUBLIC, "execute", eeDesc, null, null);
++            mv.visitVarInsn(Opcodes.ALOAD, 1);
++            mv.visitTypeInsn(Opcodes.CHECKCAST, Type.getInternalName(listenerClass));
++            mv.visitVarInsn(Opcodes.ALOAD, 2);
++            mv.visitTypeInsn(Opcodes.CHECKCAST, Type.getInternalName(eventClass));
++            int inv = listenerClass.isInterface() ? Opcodes.INVOKEINTERFACE : Opcodes.INVOKEVIRTUAL;
++            mv.visitMethodInsn(inv, Type.getInternalName(listenerClass), handlerMethod.getName(), Type.getMethodDescriptor(handlerMethod), listenerClass.isInterface());
++            if(handlerMethod.getReturnType() != void.class) {
++                mv.visitInsn(Opcodes.POP);
```
Janmm14: needs pop2 for long/double

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
```diff
@@ -0,0 +1,258 @@
...
++import java.lang.reflect.Modifier;
++
++/**
++ * @author Mark Vainomaa
++ */
++public interface EventExecutor {
++    void execute(Object listener, Object event);
++
++    static EventExecutor create(Method method) {
++        if (Modifier.isPublic(method.getModifiers())) {
```
Janmm14: does magicaccessor class still exist? if yes, we could do nonpublic calls also with asm

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
```diff
@@ -0,0 +1,258 @@
...
++
++import java.lang.invoke.MethodHandle;
++import java.lang.invoke.MethodHandles;
++import java.lang.reflect.Method;
++import java.lang.reflect.Modifier;
++
++/**
++ * @author Mark Vainomaa
++ */
++class MethodHandleEventExecutor implements EventExecutor {
```
Janmm14: i think there were huge discussions about method handle vs normal reflection and i heard that methodhandles are in the end slower than normal reflection, as the jvm optimizes normal reflection also to asm at some point

Janmm14:
This was part of waterfall in the past, someone should dig up removal reasons.

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
mikroskeem: Can't access magic stuff past Java 9 iirc, wouldn't risk with that.

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
mikroskeem: Yeeah maybe indeed.

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
Janmm14: Actually I don't think the Modifier will ever be nonpublic, as on register it only registers public methods or sth like that

mikroskeem: Will check it later and make changes.

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
mikroskeem: Does EventBus even consider methods with non-void returns as event handlers? Then could simplify that logic and remove POP entirely.

Janmm14: EventBus does not check return value of `@EventHandler` methods: https://github.com/SpigotMC/BungeeCord/blob/b85df4f2a19222bf96b9143de38438a95911eec4/event/src/main/java/net/md_5/bungee/event/EventBus.java#L61

mikroskeem:
Archaeological excavations resulted with this: <https://github.com/PaperMC/Waterfall/commit/80bed7979d707563283635998cc326f6aa0ad5b0#commitcomment-18806881>

tl; dr; issue with Techcable's Event4J library, which is not an issue here anymore as this is more raw solution.

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
mikroskeem: Where is the facepalm emoji in GitHub when I need it.

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
mikroskeem: register calls findHandlers, which in turn calls `listener.getClass().getDeclaredMethods()`, so we still end up with non-public methods as I recalled correctly.

Janmm14: bungee does call getDeclaredMethods but does never call setAccessible(true), so it would fail calling a nonpublic eventhandler

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
mikroskeem: In case of Bungee not executing non-public methods (either intentionally or not intentionally), then this event executor implementation becomes irrelevant.

mikroskeem:
Made requested changes.

Leymooo:
> This was part of waterfall in the past, someone should dig up removal reasons.

https://github.com/PaperMC/Waterfall/issues/70
https://github.com/PaperMC/Waterfall/issues/84

BungeeCord-Patches/0056-Use-ASM-to-invoke-event-handlers.patch
```diff
@@ -0,0 +1,213 @@
...
+diff --git a/event/src/main/java/net/md_5/bungee/event/EventBus.java b/event/src/main/java/net/md_5/bungee/event/EventBus.java
+index d7c3b90d..2aa4941d 100644
+--- a/event/src/main/java/net/md_5/bungee/event/EventBus.java
++++ b/event/src/main/java/net/md_5/bungee/event/EventBus.java
+@@ -51,10 +51,10 @@ public class EventBus
+                 } catch ( IllegalArgumentException ex )
+                 {
+                     throw new Error( "Method rejected target/argument: " + event, ex );
+-                } catch ( InvocationTargetException ex )
++                } catch ( Exception ex ) // Waterfall - Use ASM to invoke event handlers
```
mikroskeem, line 177: Though this makes me think. What happens when `Method#invoke(...)` throws `Throwable` instead of an `Exception`? Javadoc tells me only about [exception](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html#invoke-java.lang.Object-java.lang.Object...-).

This needs to be verified, if Throwable was passed on wrapped in InvocationTargetException, then I need to change this catch to handle Throwable instead.

xxDark:
There are still some things that can be improved by dropping `CHECKCAST` instructions when possible (see `jdk.internal.reflect.MagicAccessorImpl/sun.reflect.MagicAccessorImpl`)

Janmm14:
why got this closed?
@xxDark I don't think that normal java classes can leave out CHECKCAST, it'd get a verify error likely on class load

Hyronymos-Old:
I guess he closed all his PRs because they didnt got merged after a month and because he is active contributing to velocity now

mikroskeem:
>why got this closed?

Low interest in implementing this. If someone else wants to do this, feel free to pick up the leftovers.

xxDark:
> 
> 
> why got this closed?
> @xxDark I don't think that normal java classes can leave out CHECKCAST, it'd get a verify error likely on class load

Doing that without any problems. Plus it allows me to invoke private listener's methods

PR #503, opened 2020-05-08
Add missing permission nodes

mikroskeem:
Allows to put /bungee and /perms behind permission nodes


Janmm14:
not a fan of adding a permission node to /bungee - all it provides is security by obscurity while making waterfall less popular

mikroskeem:
I mean people can already do that with `disabled_commands`, so if they really wanted security by obscurity then they can have it.

Janmm14:
But if you add permission node, default behaviour changes.

mikroskeem:
Sure, might reconsider dropping /bungee from this patch. What about /perms? In my opinion it's pretty useless if you have custom permissions plugin (handling permission checks via an event)

PR #515, opened 2020-06-06
Reset max health attribute on server switch

kennytv:
Attributes aren't reset, so the player sees the wrong amount of hearts when switching away from a server that changed the default value to a server without the change (since default attributes are not sent to the client). All of the other attributes either don't affect the client itself or are reset due to non-default values for players, so only the max health attribute is important.

### DIFF
--- a/BungeeCord-Patches/0056-Reset-max-health-on-server-switch.patch
+++ b/BungeeCord-Patches/0056-Reset-max-health-on-server-switch.patch
@@ -1,0 +1,190 @@
+From 37e73a01d236c2436cc615778fd1393fa2d9f57d Mon Sep 17 00:00:00 2001
+From: KennyTV <kennytv@t-online.de>
+Date: Sat, 6 Jun 2020 20:23:18 +0200
+Subject: [PATCH] Reset max health on server switch
+
+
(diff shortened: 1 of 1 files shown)


Xernium:
While we are on this subject, shouldn't JoinGame be resent if the next server is != the same mode (hardcore or not hardcore) ?

BungeeCord-Patches/0056-Reset-max-health-on-server-switch.patch
```diff
@@ -0,0 +1,190 @@
...
++
++        private UUID uuid;
++        private double amount;
++        private byte operation;
++    }
++
++    public static EntityAttributes createDefaultHealth(int entityId, int protocolVersion)
++    {
++        EntityAttributes packet = new EntityAttributes( entityId );
++        String key = "generic.maxHealth"; //TODO Differentiate for 1.16: minecraft:generic.max_health
```
Janmm14, line 170: I think that todo should be solved before merge. shouldn't be hard as the method already gets the protocolVersion.

Janmm14:
> While we are on this subject, shouldn't JoinGame be resent if the next server is != the same mode (hardcore or not hardcore) ?
> @Xernium 

This is already done if you enable the option "disableEntityMetadataRewrite" in Waterfall.

It should also be checked if that does reset the client's max health, in that caser this patch would need additional ifs so we don't send it if its not actually needed.

PR #524, opened 2020-06-30
Add the MessageRaw channel

Proximyst:
Make sure to drop if/when SpigotMC/BungeeCord#2875 is merged.

This was tested using BungeeCord chat components, but adventure probably won't err.


Proximyst:
This was merged into BungeeCord first.

electronicboy:
We often don't merge stuff PR'd to both repos unless it becomes clear there is little interest from upstream, in cases where you do PR to both, it would be handy to PR without // Waterfall, etc, that way the merge resolution is dealt with automatically and treat it as a temp patch we're likely going to drop

PR #550, opened 2020-07-21
Allow customisation of the Mojang auth url

theminecoder:
I have run into situations before where the number of people connections to a proxy at once can sometimes overwhelm certain systems and cause rate limiting from mojangs end.

 One of the teams I work for have developed an in house solution to this to load balance Mojang login requests over our internal infrastructure to combat this (which may be open sourced at some point), but to use it we are required to modify this url each time with a custom build.

To keep updating simple (this is like the only thing we have in out fork) and because I'm sure that I am not the first person to run into this I figured it would make sense to be an official patch.


electronicboy:
Being perfectly honest, I can understand the desire for this, the nature of just slapping it in a config option is where I eer a bit

Janmm14:
I think the vanilla way for this is to  use a proxyselector

JRoy:
Agree with cat on this, should be a command-line option/property imo. I can see users mistakenly touching that in the config. 

Also, why aren't you just using `%s` in the URL so you can use `String#format`. I don't understand the need for your own variables here.

theminecoder:
Have moved this to a system property and switched to `String#format`.

electronicboy (approved):

JRoy (approved):

theminecoder:
bump ;)

PR #561, opened 2020-08-12
1.16.2 Update

Xernium:
@electronicboy did I miss something? Should I be changing the pom version? 
Feel free to add edits


Proximyst (approved):

riku6460 (changes requested):
Need to update the EntityRemoveEffect packet ID

Xernium:
@riku6460 That one slipped through. Please evaluate again, thanks

electronicboy (approved):

PR #569, opened 2020-09-27
Fix some messages

riku6460:


Proximyst (changes requested):
Using only one quotation mark will render it unused. `''` becomes `'`, and `'` becomes a literal and is not present in the resulting string.

This means that e.g. `no_server_permission` will now do `You dont` instead of `You don't` in the message the player sees.

riku6460:
Waterfall build 379 outputs the following because of this. https://github.com/PaperMC/Waterfall/blob/7b4d6e99a98216175470bdb3c0f8c4cc3563a78b/BungeeCord-Patches/0016-Micro-optimizations.patch#L18-L21

Proximyst:
Could've sworn that was how properties were (properly hehe) loaded. In that case, at least do it to every message with this issue.

riku6460:
I think already fixed all messages.

Janmm14:
The correct solution would be to make the no-argument message getter replace `''` with `'`.
This way waterfall has to maintain fewer changes while staying 100% compatible with bungeecord's message properties format.

And the two `''` have to stay in messages which take argument, as for those, MessageFormat is still used. So the current optimization creates confusing, inconsistent behaviour and should be corrected.

PR #585, opened 2020-11-24
added simple Dockerfile

MrDienns:
Perhaps someone will find this useful. Personally running on Kubernetes, so I need Docker images for everything. Figured I might as well share it here if anyone's interested. If this ends up being accepted, ideally you'd also setup a Github Action (or similar) to automatically push this to Dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM openjdk:11
+
+# Define the environment variables
+ENV USER=container
+ENV SERVER_DIRECTORY="/home/container"
+
+ENV JAVA_HEAP_SIZE="4G"
+ENV JAVA_ARGUMENTS="-Xmx${JAVA_HEAP_SIZE} -Xms${JAVA_HEAP_SIZE}"
+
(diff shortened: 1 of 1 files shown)


electronicboy:
Not too sure I really want this on the basis of "now we gotta maintain a docker file", at the very least, this is flawed as it doesn't even copy the modules into place either, and 4G of heap is excessive for the proxy for 99% of people, most memory churning on a supported platform is off-heap

MrDienns:
Module copying can be added. The 4G is literally an environment variable; adjust as needed (although the default can be lowered). Anyone who runs a Docker based infrastructure neneds Docker images, so I figured this could help some people.

lus:
I just want to mention that there are actively community-maintained Dockerfiles for Waterfall (https://github.com/itzg/docker-bungeecord). As @electronicboy said this would allow focusing on Waterfall itself more in this repository and not having the struggle to maintain an unneccessary Dockerfile additionally.

PR #597, opened 2021-01-21
Limit login request data to 16 bytes

### LINKED ISSUE #594: mysterious packets made bungee crash
Hey developers,

recently a mysterious package from an brazilian IP occured on my server which made it crash. Is there any way to prevent this from happening, so that those packets are automatically rejected?
[bungee-package.pdf](https://github.com/PaperMC/Waterfall/files/5825284/bungee-package.pdf)

i had to put it into a pdf because it's almost 20 din a 4 pages long ...

Thanks in advance,
Cris

hochbaum:
When decoding the data of the LoginRequest packet, up to 32767 bytes can be read although usernames should not be that huge. This commit _might_ prevent issue #594 from happening.


electronicboy:
Thanks for the PR, had been meaning to look into this but tripped into something else

Janmm14:
Patch got included in upstream.

Issue #600, opened 2021-01-26
"Overflow in packet detected" - Max 16 chars string limit not wholly fixed?


freak-dk:
Hi

I saw [this issue](https://github.com/PaperMC/Waterfall/issues/594) on a new exploit and updated my Waterfall Proxy so we're running the newest build; Waterfall-Bootstrap:1.16-R0.4-SNAPSHOT:2eae08f:393

Today we were hit by (I believe?) the same attack, causing some interruption.
At first this is spammed in the console; https://pastebin.com/c28c1kna
Then this: https://pastebin.com/jwMsncRt
And finally Session server times out, because of too many requests I guess; https://pastebin.com/tmkTuQEX

Obviously fake usernames and IPs.

So basically almost nothing came of this but when connection to the session server timed out a number of legitimate users were not able to be authenticated for a few minutes.
Seems like the fix is 'working' but still trying to authenticate the fake users?

Anything I can do about this / is this something you need to work out?

Thank your for your efforts!

Janmm14:
obviously not fake ips, as it is tcp connection.

freak-dk:
"Obviously" the **usernames** are fake... 
The IPs are proxys/vpn, not someones home connection is what I meant :)

Xernium:
Some more mitigations regarding this are being worked on. For now have some patience please.

freak-dk:
> Some more mitigations regarding this are being worked on. For now have some patience please.

Yes of course.
I wasn't sure if this was 'my' problem or something you needed to work on 👍 

Is there anything I can do about this here and now though?
It's quite the issue for me atm. with someone actively exploiting this on my proxy.

PR #603, opened 2021-01-31
Additional DoS Mitigations for the login sequence

Xernium:
Recently more overload DoS attacks have been observed on Minecraft proxies.
@astei mitigated these types of attacks on the Velocity project by introducing an upfront check for packet sizes
The commit in question can be found [here](https://github.com/VelocityPowered/Velocity/commit/5ceac16a821ea35572ff11412ace8929fd06e278)

This patch aims to bring the same mitigations found in Velocity to Waterfall.
Additionally, this patch contains some additional mitigations by limiting the generation of stacktraces by using cached exceptions instead.

The launch-flag `-Dwaterfall.packet-decode-logging=true` was also introduced to disable the usage of these cached exceptions

While these changes are tested, there might be some details that were missed, so reviews and testing of this PR would be greatly appreciated.

Should mitigate: https://github.com/PaperMC/Waterfall/issues/600


electronicboy (approved):

McRageRecoded:
hi :)

ghost:
@McRageRecoded crybaby

Issue #607, opened 2021-02-09
Further DoS fixes needed


OpenSource03:
Hello there.

The recent DoS fixes did not really do much (at least in my case). I have a sample of the java application used for sending attacks. For security reasons, I uploaded it [here](https://workupload.com/file/nHVqbYRLp5y) with a password. A developer/staff member can request me to remove the password should you wish for everyone to be able to grab this tool. Otherwise, you can contact me on Discord (OpenSource#2021) to get the file password.

Launch attack command: `java -jar jarname.jar host=127.0.0.1 port=25577 proxiesFile=socks4.txt threads=10000 attackTime=320 exploit=utf16_join` **(it's important to keep exploit method utf16_join, all else is optional).**

I hope there is something more you can do about it. This tool is able to overload any network, no matter the Bot/DDoS protection being used.

Thanks!

electronicboy:
Being personally honest, there is only so much we can do about DDoS attacks, they take advantage of the fundamental nature of software exposed to the internet: they accept connections.

We can deal with some aspects, e.g. reducing the impact of eronious data, but, large number of connections being opened to one machine, that's out of the bounds of what waterfall can sanely defend against. Configure your firewall and look at proper load balancing to mitigate these things, such is the modern internet, this is the technique which services like CF basically take advantage of, because it's the only way to deal with a large number of connections.

Janmm14:
@electronicboy

I analysed a pcap dump from him and based on that, bad connections are just not closed in a reasonable timeframe, like at max 0.5 seconds. They are not even closed in the full length of 2 seconds of his pcap dump. This happens because bungeecord is still waiting for the "Login Start" packet (the packet with the name field) to finish over the wire, before handling it and noticing that its too big. When this happens with thousands of connections, it gets overloaded.

The problem is the Varint21FrameDecoder.

The Varint21FrameDecoder needs to be copied for the direction Client -> Bungee to add very restrictive limits.

The copied and renamed LimitedVarint21FrameDecoder should the limit maximum packet sizes of the first 4 packets of each connection drastically (they are handshake -> login start -> encryption response -> brand plugin message or hand shake -> server list status request).

If people want to abuse the greedyness of the frame decoder further, it requires more heavy clientside computations (encryption) and a valid minecraft login to keep connections open for longer afterwards.

Further possible restrictions to apply:
- Frame decoder: Maximum fragmentation count for the first couple packets. 
- Stricter timeout for the first couple packets
[...]

PR #609, opened 2021-02-09
Improve additional DoS mitigations

Janmm14:
#607 

Check the first 4 packet lengths already in Varint21FrameDecoder, Handshake packet got minimum and maximum limits as well.
Also time out connections which were not able to send a completed packet within the limit. This might improve recovery after attacks.

Test jar: https://ci.janmm14.de/job/public%7Emcjava%7Ewaterfall-pr/

### DIFF
--- a/BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
+++ b/BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
@@ -1,4 +1,4 @@
-From 0fd0bf29d77463349c44fc92f9ccdcae63af89db Mon Sep 17 00:00:00 2001
+From 009157dcf9d6f447ff1c0fb399e835c454b1a8bf Mon Sep 17 00:00:00 2001
 From: "Five (Xer)" <admin@fivepb.me>
 Date: Sat, 30 Jan 2021 18:04:14 +0100
 Subject: [PATCH] Additional DoS mitigations
@@ -6,6 +6,11 @@ Subject: [PATCH] Additional DoS mitigations
 Some stricter length checks and cached exceptions to withstand more illegitimate connections
 Courtesy of Tux and the Velocity Contributors. See:
 https://github.com/VelocityPowered/Velocity/commit/5ceac16a821ea35572ff11412ace8929fd06e278
+
+Improved by Janmm14:
+Check the first 4 packet lengths already in Varint21FrameDecoder, Handshake packet got minimum and maximum limits as well.
+
+Co-authored-by: Janmm14 <gitconfig1@janmm14.de>
 
 diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java b/protocol/src/main/java/net/md_5/bungee/protocol/DefinedPacket.java
 index d10cf2ed..31a95495 100644
@@ -165,8 +170,40 @@ index 388f6cdb..53575ce0 100644
 +    }
 +    // Waterfall end
  }
+diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/packet/Handshake.java b/protocol/src/main/java/net/md_5/bungee/protocol/packet/Handshake.java
+index 70934a5b..b0a25b59 100644
+--- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/Handshake.java
++++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/Handshake.java
+@@ -7,6 +7,7 @@ import lombok.EqualsAndHashCode;
+ import lombok.NoArgsConstructor;
+ import net.md_5.bungee.protocol.AbstractPacketHandler;
+ import net.md_5.bungee.protocol.DefinedPacket;
++import net.md_5.bungee.protocol.ProtocolConstants;
+ 
+ @Data
+ @NoArgsConstructor
+@@ -43,4 +44,19 @@ public class Handshake extends DefinedPacket
+     {
+         handler.handle( this );
+     }
++
++    // Waterfall start: Additional DoS mitigations
++    // newer protocol limits host string to 255 chars (also reasonable for older clients) + allow \0FML\0 afterwards
++    public static final int EXPECTED_MAX_LENGTH = 5 + ((255 * 4 + 3) + 5) + 2 + 1;
++
++    @Override
++    public int expectedMaxLength(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion) {
++        return EXPECTED_MAX_LENGTH;
++    }
++
++    @Override
++    public int expectedMinLength(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion) {
++        return 1 + 1 + 2 + 1;
++    }
++    // Waterfall end
+ }
 diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginRequest.java b/protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginRequest.java
-index a691f962..a4c0a6a5 100644
+index a691f962..990d9a88 100644
 --- a/protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginRequest.java
 +++ b/protocol/src/main/java/net/md_5/bungee/protocol/packet/LoginRequest.java
 @@ -11,6 +11,7 @@ import lombok.EqualsAndHashCode;
@@ -177,16 +214,18 @@ index a691f962..a4c0a6a5 100644
  
  @Data
  @NoArgsConstructor
-@@ -38,4 +39,12 @@ public class LoginRequest extends DefinedPacket
+@@ -38,4 +39,14 @@ public class LoginRequest extends DefinedPacket
      {
          handler.handle( this );
      }
 +
 +    // Waterfall start: Additional DoS mitigations, courtesy of Velocity
++    public static final int EXPECTED_MAX_LENGTH = 1 + (16 * 4);
++
 +    public int expectedMaxLength(ByteBuf buf, ProtocolConstants.Direction direction, int protocolVersion) {
 +        // Accommodate the rare (but likely malicious) use of UTF-8 usernames, since it is technically
 +        // legal on the protocol level.
-+        return 1 + (16 * 4);
++        return EXPECTED_MAX_LENGTH;
 +    }
 +    // Waterfall end
  }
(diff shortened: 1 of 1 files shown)


Janmm14:
Should we add the channel not active checks from `Varint21FrameDecoder` and others also to `PacketDecompressor` and `CipherDecoder`?

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -49,6 +55,129 @@ index d10cf2ed..31a95495 100644
 +    }
 +    // Waterfall end
  }
+diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/LimitedVarint21FrameDecoder.java b/protocol/src/main/java/net/md_5/bungee/protocol/LimitedVarint21FrameDecoder.java
```
Xernium: Please don't copy classes just like that. Instead you could store a Boolean to materialise the limit(s) when the decoder is conceptualised.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -49,6 +55,129 @@ index d10cf2ed..31a95495 100644
...
++                        case 2:
++                            maxLength = EncryptionResponse.EXPECTED_LENGTH + 2;
++                            break;
++                        case 3:
++                            // minecraft:brand plugin channel message 
++                            // minecraft:brand (16 bytes) followed by a 400 char long string should never be reached
++                            maxLength = 16 + (400 * 4 + 3); 
++                            break;
++                    }
++                    if (length > maxLength) {
```
Xernium: Instead of doing this, put it together with the same length sanity checks that are already in this patch and move it to a papermc package class

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -240,6 +406,66 @@ index 738f0c92..ec33d337 100644
...
++import net.md_5.bungee.protocol.LimitedVarint21FrameDecoder;
+ import net.md_5.bungee.protocol.MinecraftDecoder;
+ import net.md_5.bungee.protocol.MinecraftEncoder;
+ import net.md_5.bungee.protocol.Protocol;
+@@ -82,7 +83,7 @@ public class PipelineUtils
+ 
+ 
+             try {
+-            BASE.initChannel( ch );
++            LIMITED_BASE.initChannel( ch ); // Waterfall - Use LIMITED_BASE (Additional DoS mitigations)
```
Xernium, line 388: Again, use an additional boolean bit instead

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -240,6 +406,66 @@ index 738f0c92..ec33d337 100644
...
+             } catch (Exception e) {
+                 e.printStackTrace();
+                 ch.close();
+@@ -103,10 +104,12 @@ public class PipelineUtils
+             BungeeCord.getInstance().getPluginManager().callEvent(connectionInitEvent);
+         }
+     };
+-    public static final Base BASE = new Base();
++    public static final Base BASE = new Base(false); // Waterfall - Base class takes parameter now (Additional DoS mitigations)
++    public static final Base LIMITED_BASE = new Base(true); // Waterfall - Add LIMITED_BASE for connections form clients (Additional DoS mitigations)
```
Xernium, line 396: Unnecessarily duplicated, use a boolean bit in that instead

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -240,6 +406,66 @@ index 738f0c92..ec33d337 100644
...
+         @Override
+         public void initChannel(Channel ch) throws Exception
+@@ -191,8 +201,9 @@ public class PipelineUtils
+             ch.config().setWriteBufferWaterMark( MARK );
+ 
+             ch.pipeline().addLast( TIMEOUT_HANDLER, new ReadTimeoutHandler( BungeeCord.getInstance().config.getTimeout(), TimeUnit.MILLISECONDS ) );
+-            ch.pipeline().addLast( FRAME_DECODER, new Varint21FrameDecoder() );
++            ch.pipeline().addLast( FRAME_DECODER, fromClient ? new LimitedVarint21FrameDecoder() : new Varint21FrameDecoder() ); // Waterfall - limit decoding length for client packets (Additional DoS mitigations)
+             ch.pipeline().addLast( FRAME_PREPENDER, framePrepender );
++            ch.pipeline().addLast( FRAMED_TIMEOUT_HANDLER, new ReadTimeoutHandler( BungeeCord.getInstance().config.getTimeout(), TimeUnit.MILLISECONDS ) ); // Waterfall - limit time we wait on complete (non-framed) packets
```
Xernium: This is outside the scope of this and potentially creates a race condition

Xernium (changes requested):

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: If I want to edit the initChannel method signature, I'd have to remove it's `extends ChannelInitializer<Channel>`.
I thought I'd be worse to edit class hierarchy, that is why I did it with the constructor parameter.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: I already put it together through the max length constants above. I don't see an existing fitting methods for this in this patch.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: Ok, once the other points are cleared, I'll remove this one.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -49,6 +55,129 @@ index d10cf2ed..31a95495 100644
...
++                    int maxLength = 2097151; // max length of 21-bit varint
++                    switch (packetCount) {
++                        case 0:
++                            maxLength = Handshake.EXPECTED_MAX_LENGTH + 2;
++                            break;
++                        case 1:
++                            maxLength = LoginRequest.EXPECTED_MAX_LENGTH + 1;
++                            break;
++                        case 2:
++                            maxLength = EncryptionResponse.EXPECTED_LENGTH + 2;
```
Leymooo: offline mode does not use encryprion stuff.

Janmm14: Oh yeah forgot about that. Will add a check for that.

Janmm14: I've put in higher limits to account for offline mode.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: Removed

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: Please elaborate what you mean with boolean bit. I think this change is reasonable. I've also just further reduced the impact of the change by providing a default constructor for the Base class.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: I've done some git magic and we have the patch now describing it as a file copy.
If that is not clean enough for you, I can happily add a constructor boolean to enable/disable limiting behaviour in Varint21FrameDecoder.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: Moved LimitedVarint21FrameDecoder to a papermc package. For the other part of your message, please explain in more detail what you mean.

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
Janmm14: @Xernium Can you please elaborate why this would create a race condition?

ghost:
merge?

BungeeCord-Patches/0060-Additional-DoS-mitigations.patch
```diff
@@ -163,10 +168,42 @@ index 388f6cdb..53575ce0 100644
...
+ @Data
+ @NoArgsConstructor
+@@ -43,4 +44,19 @@ public class Handshake extends DefinedPacket
+     {
+         handler.handle( this );
+     }
++
++    // Waterfall start: Additional DoS mitigations
++    // newer protocol limits host string to 255 chars (also reasonable for older clients) + allow \0FML\0 afterwards
++    public static final int EXPECTED_MAX_LENGTH = 5 + ((255 * 4 + 3) + 5) + 2 + 1;
```
astei, line 192: This will have to be increased or removed - Minecraft-specific DDoS protection services such as TCPShield and Cosmic Guard like to stuff extra data in the handshake address field. I did implement this in Velocity but due to the aforementioned issue I had to revert it.

Janmm14: I will put in a property switch to set the maximum size for that packet then.

I am sure there are many waterfall users not using such ddos protections, they need the limit.

paulzhng: 255 characters is plenty for our (TCPShield) purposes.

Janmm14: It is also multiplied by 4 for the heaviest utf-8 chars taking 4 bytes, so its actually much more, if you just use 1 or 2 byte utf-8 chars only.
Then I guess it is actually not needed to allow editing this value.

ghost:
any update?

electronicboy:
I'm ill, so my capacity to throw mental energy into stuff like this is limited and so I value the commentary made by others in the community with experience

Please don't bump PRs unless you have something to add, have you tested it? does it work? etc

Janmm14:
> Please don't bump PRs unless you have something to add, have you tested it? does it work? etc

In the issue, the issue creator already wrote that it does fix the exploit.

I think the changes done in this PR are quite straightforward and it should get merged soon.

I'm happy to assist when problems arise due to this change.

One thing that might change is the way the software reacts when another bungee with ip/uuid/skin forwarding enabled is pointed to a waterfall instance with this patch. As the skin data might be longer than the limit we have for the server name (255*4=1020 ascii chars), it might get disconnected without the kick message bungee put in place for this scenario.
But I think that this is nothing we really should worry about.

electronicboy:
a singular (set of) test(s) in a very specific environment is not really a good means of knowing that such a patch is ready to pull, this patch feels too heavily implementation detailed for me in some areas, has this been tested with any mod packs or anything like that for example?

Janmm14:
Ah yeah I forgot the stuff which mods might do. My bad.

I guess we'll have to wait for some people who use my Waterfall PR in combination with modpacks to see if forge / any modpack sends long data in the first 4 packets to test it out.
While I am happy to assist with potential errors arising due to mods (like increasing the already big limits or adding some config option to only check the first 3 instead of 4 packets or so), I don't feel like digging into modpack server & client installations in different versions etc.

OpenSource03:
I've been using this build for a long time now, and I can confirm that on a normal Waterfall with a moderate amount of players, some plugins, and continuous bot/DDoS attacks (I run a JH AntiBot test server), it performed really well. No issues at all. This is a **MUST** to be merged. Default Bungee implementation is not only allowing for a network overload, but also getting overloaded itself, consuming an unlimited amount of RAM during the attack, which does not get cleared automatically, even after the attack is stopped, while also overloading CPU.

Janmm14:
Rebased onto latest waterfall changes, mainly for my CI.

OpenSource03:
Such a important pull request is not gonna be merged any time soon. Multiple large hosting providers have tried to "mitigate" it with firewall, but idk how do you expect to block legit Minecraft connections (bots) with firewall... I've been using this for months and there are just no issues. Why waiting more? Reject it or accept it...

electronicboy:
Because I'm ill and do not have the time to answer the questions I've asked myself, and apparently nobody else has cared to put in the effort either.

if you want to help get this pulled, please, answer the question for us, go test in supported modded environments.

electronicboy:
I'm inclined to close this pr, my health is generally shot and the attitude/demands I get over this PR far exceeds the people willing to actually answer and address the concerns I have. For a free project I don't get a cent for, I'm not all that inclined to throw much mental energy into such a questionable PR which nobody else wants to throw the effort into either

astei:
> Such a important pull request is not gonna be merged any time soon. Multiple large hosting providers have tried to "mitigate" it with firewall, but idk how do you expect to block legit Minecraft connections (bots) with firewall... I've been using this for months and there are just no issues. Why waiting more? Reject it or accept it...

Your condescending attitude is unhelpful. Leave it at the door.

Proximyst (changes requested):
This looks extremely janky. Why do we not parse *what* the packet is first, then to check expected packet length which is already defined in their classes?

Janmm14:
>  _@Proximyst wrote:_
> This looks extremely janky. Why do we not parse _what_ the packet is first, then to check expected packet length which is already defined in their classes?

That is already done by the current version of the patch this PR modifies.

The attack sends a large packet size and sends the data very slowly but frequently enough to not trigger a timeout (as the timeout handler is before the Varint21FrameDecoder). This causes the connections to stay open for a long time without effort from the attacker, while the packet never leaves the Varint21FrameDeocder stage of the netty pipeline (causing no additional checks to happen).

It also doesn't show up in the logs as new connections are first logged after decoding the handhsake packet.

Originally I additionally added a second timeout handler after the Varint21FrameDecoder to catch all, also the more sophisticated attacks eventually, but it was claimed that it might cause some race conditions (without further explanation), so I removed it from this PR again. See https://github.com/PaperMC/Waterfall/pull/609#discussion_r573906817

So what is left is limiting the first 4 (I chose this number as the first 4 packet types minecraft sends are kinda well-defined) packets from connecting clients - especially for online-mode bungeecords this prevents such attacks as that includes the EncryptionResponse packet and that requires some effort from the client and a valid mc login to not get kicked at this stage.
The idea was to prevent the 30 second timeout waiting for clearly illegal incoming traffic.

The current solution is a little hacky, but works well with vanilla minecraft clients.

As Waterfall aims for modded compatibility as well, this is not merged yet. In theory, mods might send other, potentially bigger packets during the login sequence, causing them to get kicked falsely.

I hoped @Xernium to answer why a second ReadTimeoutHandler would potentially create a race condition, he never did though.

To help **electronicboy** relax, I will propse the second ReadTimeoutHandler solution after the Varint212FrameDecoder to BungeeCord as a potential alternative solution to various kinds of attacks.

electronicboy:
the server already has a thing to limit how long a client can spend on LOGIN, iirc, maybe worth looking at how long that allowed and such

Janmm14:
https://github.com/SpigotMC/BungeeCord/pull/3066

Janmm14:
Closing, as this changed is deemed too risky.

PR #644, opened 2021-05-21
Add a command to view all proxy plugins

Findus642:
Following from the previous PR before git decided to break, this command simply allows a server owner, developer, etc.. see what plugins are active on the proxy easily

Since the last PR:

- Proper use of BungeeCord Chat API
- Relocated to ```io.github.waterfallmc``` package
- Any previous changes that were suggested have also been carried over

### DIFF
--- a/BungeeCord-Patches/0062-Add-command-to-view-all-proxy-plugins.patch
+++ b/BungeeCord-Patches/0062-Add-command-to-view-all-proxy-plugins.patch
@@ -1,0 +1,79 @@
+From 481b64d2bac4530c108e0116666c91db2d6526a5 Mon Sep 17 00:00:00 2001
+From: MrFishCakes <fishcake007@outlook.com>
+Date: Fri, 21 May 2021 17:32:13 +0100
+Subject: [PATCH] Add command to view all proxy plugins
+
+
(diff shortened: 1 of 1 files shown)


BungeeCord-Patches/0062-Add-command-to-view-all-proxy-plugins.patch
```diff
@@ -0,0 +1,79 @@
...
++        final ComponentBuilder builder = new ComponentBuilder();
++
++        builder.append("Plugins (" + plugins.size() + "): ");
++        boolean firstIteration = true;
++
++        for (Plugin plugin : plugins) {
++            if (firstIteration) {
++                firstIteration = false;
++            } else {
++                builder.append(ChatColor.WHITE.toString()).append(", ");
```
Janmm14: You set the color not by prepending the ChatColor toString, but by calling .color(...) after append

Findus642: Now fixed.

electronicboy:
Thanks for the PR, However, adding new commands to the proxy is always this really weird area given how it will blindly interecept commands even if they're defined by the server, for this reason, I'd require that either the name was stupidly unlikely to conflict with something else, or served a substantiative purpose as to why it was added, and I'm not sure that this PR really serves either one

PR #649, opened 2021-06-02
Added an api method to unload plugins

abhithedev200:
so finally a api method to unload plugins fully at runtime. requested by an user on waterfall discord

### DIFF
--- a/BungeeCord-Patches/0062-Added-an-api-method-to-unload-plugins.patch
+++ b/BungeeCord-Patches/0062-Added-an-api-method-to-unload-plugins.patch
@@ -1,0 +1,83 @@
+From 57c17912d44f13473ef78771ada30eb6cedb4d48 Mon Sep 17 00:00:00 2001
+From: abhiram <abhithegamer2019@gmail.com>
+Date: Thu, 3 Jun 2021 01:26:53 +0530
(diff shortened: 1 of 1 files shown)


Xernium:
This functionality isn’t something that should be supported in any scenario. Un-or-reloading of plugins at runtime is never completely clean and leads to more issues than it solves. I hope that the paper team shares my notion to reject this

astei:
There are many, many plugins that do not tolerate being cleaned up in this fashion. This was something we learned from Bukkit/Spigot and it wasn't a good idea then, and it isn't a good idea now. Major plugins like ViaVersion already provide no guarantee of support if unloaded at run time.

We should not do this.

abhithedev200:
well i would say its developers responsibility unload correct plugins. like not unloading the plugin that serves dependency for other plugins. other option is to remove classloader release. so whats the better way for this?

electronicboy:
The JVM was not designed to deal with this and I see little reason for trying to support this mess given that there are already plugins out there which allow people to shoot themselves in the foot for this

abhithedev200:
true, but just calling onDisable and unregister plugins commands,listeners,tasks and removing it from loaded plugins would be less mess i suppose. please correct me if i am wrong.

PR #650, opened 2021-06-04
Enable 1.17 Release-protocol

Xernium:
I don't expect there to be more changes that will impact the network logic Waterfall needs to function, so we can have this early (hopefully)


electronicboy (approved):

momokli:
I am trying to connect to waterfall v431 using the latest snapshot (tried pre-release 1->5 and RC1+2) but I am always failing to connect:
`Outdated Server! I'm still on 1.8.x, 1.9.x, 1.10.x, 1.11.x, 1.12.x, 1.13.x, 1.14.x, 1.15.x, 1.16.x, 1.17`

Is it, that snapshot releases have to be allowed explicitly? What can I do to solve that?

electronicboy:
We don't support snapshots or pre-release builds, you'd need to add the protocol info yourself, but, there has been no changes so just need to add the protocol ID pretty much

Bixilon:
Will work, there have been no changed in the protocol anymore.

StefanH-AT:
I can confirm.

PR #660, opened 2021-06-19
Updated Upstream (BungeeCord)

linsaftw:
* Check chat for illegal chars & moved length check into the packet class

I had to fix some compilation issues related to the move of the chat message length check.


electronicboy:
Thanks for the PR however this has already been handled

PR #661, opened 2021-06-21
Cumulated recent prs from me to bungeecord

Janmm14:
- Use Java 8 lambda map access in EventBus
Improves speed as HashMap doesn't have to calculate hash multiple times.
- Speed up event calling with LambdaMetafactory BiConsumer creation
Using lookup of special class loaded by a class loader which can delegate to all PluginClassloaders.
- Replace reflection inside netty with ChannelFactory.
Thanks for pointing it out @MrIvanPlays
- Use suppliers instead of reflection for native impl generation.
Thanks for pointing it out @MrIvanPlays
- ~~Remove unused AttributeKeys in PipelineUtils~~
- Use switch in "BungeeCord" plugin message handling
- Include amount of extra bytes in exception message
Also made the message a little more similar to vanilla.


Janmm14:
Just noticed that I forgot to add Waterfall comments to this. Will fix it later (approx in 35 hours), not now.
If someone else wants this real fast in Waterfall, fell free to do it.
In general you have my permission to use my BungeeCord PRs in Waterfall when crediting me.

Janmm14:
~Waterfall vibes~~

electronicboy:
Sorry for the delay, health and all, but, I'd be somewhat inclined to ask for this to be separated into separate PRs so that we can actually go through them individually

Janmm14:
@electronicboy
They are still individual patches. Should be quite easy to go through them individually.

Janmm14:
I created PR 797 as alternative to this PR, those seeking eventbus speed improvements can find them inside IvanCord.
The other things got merged into bungee.

PR #663, opened 2021-06-25
Extended Information-forwarding support

Xernium:
This PR brings Waterfall on a forwarding-feature-level with [Velocity](https://github.com/VelocityPowered/Velocity) for better security.
Implemented:
 - Velocity modern forwarding support, courtesy of @astei and the Velocity contributors, this forwarding mode uses a 1.13 and newer feature for a more standardised and less volatile forwarding solution.
 - Bungeeguard forwarding support (natively), courtesy of @lucko this forwarding type is an extension to the standard/legacy Bungeecord forwarding protocol which adds another field/token for authentication.    

This PR is soon to be accompanied by a PR to update the PaperMC/PaperDocs with information on all of this

### DIFF
--- a/BungeeCord-Patches/0061-Rework-information-forwarding.patch
+++ b/BungeeCord-Patches/0061-Rework-information-forwarding.patch
@@ -1,4 +1,4 @@
-From ba6e0e6d2baea3310434b8d2915b0d4a0f823ee0 Mon Sep 17 00:00:00 2001
+From 08f6ce558f755e916b7a1eebff380a6f010acdcb Mon Sep 17 00:00:00 2001
 From: "Five (Xer)" <admin@fivepb.me>
 Date: Wed, 23 Jun 2021 23:10:02 +0200
 Subject: [PATCH] Rework information forwarding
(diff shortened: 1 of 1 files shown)


BungeeCord-Patches/0061-Rework-information-forwarding.patch
```diff
@@ -0,0 +1,371 @@
...
+ }
+diff --git a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+index 90031156..f1aad373 100644
+--- a/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
++++ b/api/src/main/java/net/md_5/bungee/api/plugin/PluginManager.java
+@@ -439,6 +439,14 @@ public final class PluginManager
+                         Preconditions.checkNotNull( desc.getName(), "Plugin from %s has no name", file );
+                         Preconditions.checkNotNull( desc.getMain(), "Plugin from %s has no main", file );
+ 
++                        // Waterfall start: Forwarding rework
```
electronicboy, line 94: I err on this, no idea if others have any other thoughts on this, I mean, it makes sense but I don't want this to become a norm in any form of the sense

electronicboy (approved):

BungeeCord-Patches/0061-Rework-information-forwarding.patch
lucko: If it's any help, I'd happily accept a PR in BungeeGuard that implements this the other way around :)

electronicboy: Yea, that would probs be nicer, just feels a bit "shameful" like this, and I don't wanna set the means for that oddball slippy slope of "shaming" plugins <3

electronicboy:
I kinda want this, just solely because I can't be arsed breaking my setup to test waterfall cruddery, there was a linked PR which is closed now, which I have no idea where that derives from, etc, etc; is there any interest in updating this, if Luck has no issues, in good faith, we can probably keep the message as-is

xism4:
> I kinda want this, just solely because I can't be arsed breaking my setup to test waterfall cruddery, there was a linked PR which is closed now, which I have no idea where that derives from, etc, etc; is there any interest in updating this, if Luck has no issues, in good faith, we can probably keep the message as-is

The PR it's same as here but without that excesive log and higher version support

PR #667, opened 2021-07-05
Updated Upstream (BungeeCord)

LemonCaramel:
~~Supports upcoming 1.17.1 protocol.~~

~~There is no significant protocol difference between 1.17 and 1.17.1 snapshots, so adding a protocol ID can expect normal operation.~~

~~This patch should be either removed or close(PR) when BungeeCord supports 1.17.1~~

Bungeecord 1.17.1 has been released. Switch this PR to an upstream update.


Xernium (approved):

Xernium:
@electronicboy Go now MERGE

kennytv (approved):
pufferfish

PR #669, opened 2021-07-09
Add messages file extractor

Lorenzo0111:
This pull request adds a file extractor to extract the messages.properties file outside the jar adding a way to edit messages easilly.


electronicboy:
This breaks updating the file on our end

PR #679, opened 2021-07-19
Strip hostname from InetSocketAddress

mibac138:
Recently PaperSpigot added a host check that doesn't allow hostnames
in ip addresses and if a hostname is present then it kicks the player
with a configure ip forwarding in bungee to connect error message.
By default Bungee doesn't actually include the hostname (as to include
it you need to call getHostName which triggers reverse service lookups
that add the hostname info). However, the only thing a bungee plugin
needs to do is to call addr.getHostName() in PlayerLoginEvent to
trigger this and block the player from joining any recent paper servers.
This change strips the hostname from the ip address sent to the proxied
server thus fixing this issue. Example address sent before:
`123.123.123.123.ipv4.supernova.orange.pl`
after:
`123.123.123.123`.

Context:
hi! i have a papermc 1.16.5 and bungeecord server. i updated papermc and
the server started rejecting me saying i need to configure If you wish
to use IP forwarding, please enable it in your BungeeCord config as
well but everything in config was good. so i started digging and
found out papermc added a host_check like this
Pattern HOST_PATTERN = Pattern.compile("[0-9a-f\\.:]{0,45}");
If this pattern doesn't match the hostname in
PacketHandshakeSetProtocol, the server kicks the user with the above
message. In debug I saw that my hostname looks like this:
123.123.123.123.ipv4.supernova.orange.pl. I messaged papermc about this
and they said an dns resolved ip like this should not be possible.
I started digging in bungeecord and found out an innocent plugin that
called getConnection().getAddress().getHostName() caused the address to
gain the ipv4..... suffix (getHostName does that under the hood).
I once again contacted paper devs about this and was told this is a
bungeecord issue, and its not on their side. So, long story short,
when connecting to a proxied server you need to make sure not to send
the host name

### DIFF
--- a/BungeeCord-Patches/0061-Strip-hostname-from-InetSocketAddress.patch
+++ b/BungeeCord-Patches/0061-Strip-hostname-from-InetSocketAddress.patch
@@ -1,0 +1,55 @@
+From 1b1107674563adebad11d00b6625bf8f6cbb97d1 Mon Sep 17 00:00:00 2001
+From: mibac138 <5672750+mibac138@users.noreply.github.com>
+Date: Tue, 20 Jul 2021 00:47:59 +0200
+Subject: [PATCH] Strip hostname from InetSocketAddress
+
+Recently PaperSpigot added a host check that doesn't allow hostnames
+in ip addresses and if a hostname is present then it kicks the player
+with a configure ip forwarding in bungee to connect error message.
+By default Bungee doesn't actually include the hostname (as to include
+it you need to call getHostName which triggers reverse service lookups
+that add the hostname info). However, the only thing a bungee plugin
+needs to do is to call addr.getHostName() in PlayerLoginEvent to
+trigger this and block the player from joining any recent paper servers.
+This change strips the hostname from the ip address sent to the proxied
+server thus fixing this issue. Example address sent before:
+`123.123.123.123.ipv4.supernova.orange.pl`
+after:
+`123.123.123.123`.
+
+Context:
+hi! i have a papermc 1.16.5 and bungeecord server. i updated papermc and
+the server started rejecting me saying i need to configure If you wish
+to use IP forwarding, please enable it in your BungeeCord config as
+well but everything in config was good. so i started digging and
+found out papermc added a host_check like this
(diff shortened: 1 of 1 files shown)


electronicboy (approved):

mibac138:
Many thanks for your quick response @electronicboy

kashike:
Please use the correct name for our projects - **Paper**, *not* PaperSpigot.

Timeox2k:
Is there no other way to fix this? We have a check in DiamondProtector which ensures that you can join via the hostname, but if you join via the IPv4 in Minecraft, you can be banned directly with Linux IPTables.

By removing this now, this check has been broken.

What was the point of this check?

Bot attacks resolve the IPv4 through the srv record for example and then attack the server. Since they do this automatically most of the time, whole attack websites were directly blocked.

Since you removed it now, we had to remove this check and now have less possibility to detect it.

Is there an alternative way to efficiently get the hostname as you have specified there?

electronicboy:
There is no check here, this is just ensuring that only the actual IP of the player is passed through to the backend server to cater for spigots IP host checks.

This has 0 bearing in what host/IP you connected to in the client, it's completely detached from this

Timeox2k:
You dont understand what I am saying.

I am talking about my Plugin depending on Waterfall.

I use the PlayerHandShake Event therefore I get from InetSocketAdress the getHost()

and from getHost() I check if it contains the Server Hostname if yes -> do nothing else -> blacklist

now as getHost() is removed, you broke my detection.

electronicboy:
You're not understanding what I'm saying.

getHost is not removed, the data is filtered when it is written to the handshake which is sent to the backend server, it is not removed from anywhere else, just filtered before it's passed to the server, as if something causes the players IP address to be resolved, it will cause an erroneous disconnect on the backend server.

That hostname has ZERO bearings on what is exposed in the API at all, nor is the host which is filtered relevant at all to the hostname which the client entered into their environment, it's only the clients IP address itself.

Timeox2k:
Weird, because since one of the last few Updates my Check is broken anyways. So there seems to be an Update that has broken my check. My customers started to complain since they updated Waterfall

PR #693, opened 2021-09-30
Add root/admin user warning

NoahvdAa:
This PR ports the root detection introduced in PaperMC/Paper#2432 (and re-added in PaperMC/Paper#6593) to Waterfall. The implementation used in PaperMC/Paper#6593 doesn't seem to work on Java 8, so I've used a slightly different check.

Tested on macOS and Windows 10.

### DIFF
--- a/BungeeCord-Patches/0062-Add-root-admin-user-detection.patch
+++ b/BungeeCord-Patches/0062-Add-root-admin-user-detection.patch
@@ -1,0 +1,75 @@
+From 5fb36dc51158d5d396411ac9bcf66ab7af62cfdb Mon Sep 17 00:00:00 2001
+From: Noah van der Aa <ndvdaa@gmail.com>
+Date: Thu, 30 Sep 2021 16:59:18 +0200
+Subject: [PATCH] Add root/admin user detection
+
+This patch detects whether or not the server is currently executing as a privileged user and spits out a warning.
+The warning serves as a sort-of PSA for newer server admins who don't understand the risks of running as root.
+We've seen plenty of bad/malicious plugins hit markets, and there's been a few close-calls with exploits in the past.
+Hopefully this helps mitigate some potential damage to servers, even if it is just a warning.
+
+Co-authored-by: egg82 <eggys82@gmail.com>
+
+diff --git a/api/src/main/java/io/github/waterfallmc/waterfall/utils/ServerEnvironment.java b/api/src/main/java/io/github/waterfallmc/waterfall/utils/ServerEnvironment.java
+new file mode 100644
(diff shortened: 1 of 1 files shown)


BungeeCord-Patches/0062-Add-root-admin-user-detection.patch
```diff
@@ -0,0 +1,77 @@
...
++
++import java.io.BufferedReader;
++import java.io.IOException;
++import java.io.InputStreamReader;
++
++public class ServerEnvironment {
++	private static final boolean RUNNING_AS_ROOT_OR_ADMIN;
++
++	static {
++		boolean isWindows = System.getProperty("os.name").startsWith("Windows");
```
Chew: tabs 😢 

Does this implement the fixed check from the new PR?

NoahvdAa: Java 8 (or at least in Waterfall's setup) doesn't seem to like the NTSystem/UnixSystem classes, so I used the "running a command" method from the start, so this isn't affected by the OpenJDK bug.

xism4:
> This PR ports the root detection introduced in [PaperMC/Paper#2432](https://github.com/PaperMC/Paper/pull/2432) (and re-added in [PaperMC/Paper#6593](https://github.com/PaperMC/Paper/pull/6593)) to Waterfall. The implementation used in [PaperMC/Paper#6593](https://github.com/PaperMC/Paper/pull/6593) doesn't seem to work on Java 8, so I've used a slightly different check.
> 
> Tested on macOS and Windows 10.

I think this is useless, since most people don't have access to SSH, and those who do have access already have some experience, it seems to me that adding a warning for that is redundant.

Janmm14:
> > This PR ports the root detection introduced in [PaperMC/Paper#2432](https://github.com/PaperMC/Paper/pull/2432) (and re-added in [PaperMC/Paper#6593](https://github.com/PaperMC/Paper/pull/6593)) to Waterfall. The implementation used in [PaperMC/Paper#6593](https://github.com/PaperMC/Paper/pull/6593) doesn't seem to work on Java 8, so I've used a slightly different check.
> > Tested on macOS and Windows 10.
> 
> I think this is useless, since most people don't have access to SSH, and those who do have access already have some experience, it seems to me that adding a warning for that is redundant.

It is not redundant. I believe many people running a server network do have a vserver/vps or root server rented, because they're often cheaper compared to multiple single mc servers.

Outfluencer:
I dont like the way you doing it i would use io.netty.util.internal.PlatformDependent#maybeSuperUser() instead if your reg check and creating processes, this should be enought

PR #719, opened 2021-12-20
Updated Upstream (BungeeCord)

NoahvdAa:
Upstream has released updates that appear to apply and compile correctly.
This update has not been tested by PaperMC and as with ANY update, please do your own testing

BungeeCord Changes:
b4ccdaa5 #2715: Improve BadPacketException message in MinecraftDecoder
3a116569 #3116: Do not fill in LogRecord caller data by default in slf4j wrapper
2479fab6 #3221: Use computeIfAbsent method in EventBus
51eb1ac6 Dependency upgrades
879f37f0 Upgrade to SnakeYAML 1.30 release

### DIFF
--- a/BungeeCord-Patches/0057-Additional-DoS-mitigations.patch
+++ b/BungeeCord-Patches/0057-Additional-DoS-mitigations.patch
@@ -1,4 +1,4 @@
-From 56a334d52b88a16b48046f8a4fcc8d69bb17387d Mon Sep 17 00:00:00 2001
+From a97fd919809d2284cdc6a5d68168c99bfe795d9c Mon Sep 17 00:00:00 2001
 From: "Five (Xer)" <admin@fivepb.me>
 Date: Sat, 30 Jan 2021 18:04:14 +0100
 Subject: [PATCH] Additional DoS mitigations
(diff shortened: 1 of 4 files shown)


Xernium (changes requested):
Thank you for the PR. As there are a few changes upstream that seem problematic I’m blocking this until I or another paper member have some time to verify it doesn’t break anything, please wait until then, thanks.

BungeeCord-Patches/0057-Additional-DoS-mitigations.patch
```diff
@@ -71,15 +71,16 @@ index 37d28c01..01d35f41 100644
  
                  if ( in.isReadable() )
                  {
+-                    throw new BadPacketException( "Packet " + protocol + ":" + prot.getDirection() + "/" + packetId + " (" + packet.getClass().getSimpleName() + ") larger than expected, extra bytes: " + in.readableBytes() );
```
Xernium: The patch here somehow reverts the last upstream commit, and while it applies correctly I think this should be addressed here.
This line may not be missing or no exception will ever be thrown when in debug mode

Xernium (changes requested):
Thank you for waiting, there is one small thing that needs to be changed, please see my comment

BungeeCord-Patches/0057-Additional-DoS-mitigations.patch
NoahvdAa: Thanks, fixed!

Xernium (approved):

PR #732, opened 2022-04-11
Waterfall shows to everyone ip behind backend servers - patch

MG8853:
BungeeCord-Patches/0064-Waterfall-shows-to-everyone-ip-behind-backend-server.patch

https://github.com/SpigotMC/BungeeCord/issues/3258


Arobcher:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

MG8853:
Waterfall shows to everyone ip behind backend servers if there are not online

MG8853:
i'm setting to show to ip only admin.

electronicboy:
1) upstreams changes would be pulled in an upstream merge, not manually applying changes like that
2) waterfall already prevents such info being leaked, bungee is opening it up to be displayed to admins, or a shorter message (at the cost of hiding it from server owners too)

tomcraft:
No it doesnt as there is already a patch for this. I'm not a core team member but I realy think the change from upstream is uggly and should use a permission instead of checking groups this way. 
Also this is not the proper way to update upstream.

PR #783, opened 2022-11-17
Use Velocity VarIntFrameDecoder

xism4:
Velocity provides a pretty frame decoder that could be used without any problem in Waterfall (Also works on java 8)

After debating for some time, I want to continue improving Waterfall to the best of my knowledge. Since I work with it and many people are still using it. 
And if anyone wonders why if Velocity already has it? The same question should be asked with other patches that Waterfall uses from Velocity as well.

### DIFF
--- a/BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
+++ b/BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
@@ -10,20 +10,8 @@ index c0d37142..bdeab0dc 100644
 +++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
 @@ -1,65 +1,78 @@
 +/*
-+ * Copyright (C) 2018 Velocity Contributors
-+ *
-+ * This program is free software: you can redistribute it and/or modify
-+ * it under the terms of the GNU General Public License as published by
-+ * the Free Software Foundation, either version 3 of the License, or
-+ * (at your option) any later version.
-+ *
-+ * This program is distributed in the hope that it will be useful,
-+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
-+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-+ * GNU General Public License for more details.
-+ *
-+ * You should have received a copy of the GNU General Public License
-+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
++ * Courtesy of Velocity
++ * See https://github.com/VelocityPowered/Velocity/commit/5ceac16a821ea35572ff11412ace8929fd06e278
 + */
 +
  package net.md_5.bungee.protocol;
@@ -140,20 +128,8 @@ index 00000000..2e4ccf05
 +++ b/protocol/src/main/java/net/md_5/bungee/protocol/VarintByteDecoder.java
 @@ -0,0 +1,68 @@
 +/*
-+ * Copyright (C) 2018 Velocity Contributors
-+ *
-+ * This program is free software: you can redistribute it and/or modify
-+ * it under the terms of the GNU General Public License as published by
-+ * the Free Software Foundation, either version 3 of the License, or
-+ * (at your option) any later version.
-+ *
-+ * This program is distributed in the hope that it will be useful,
-+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
-+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-+ * GNU General Public License for more details.
-+ *
-+ * You should have received a copy of the GNU General Public License
-+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
++ * Courtesy of Velocity
++ * See https://github.com/VelocityPowered/Velocity/commit/5ceac16a821ea35572ff11412ace8929fd06e278
 + */
 +
 +package net.md_5.bungee.protocol;
--- a/BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
+++ b/BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
@@ -1,0 +1,213 @@
+From dcbe861c2c63b09edc88ca7f3e42d9f9687f63af Mon Sep 17 00:00:00 2001
+From: xIsm4 <minelatinsoporte@gmail.com>
+Date: Thu, 17 Nov 2022 10:52:42 +0100
+Subject: [PATCH] Use Velocity VarintFrameDecoder
+
+
+diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
+index c0d37142..bdeab0dc 100644
+--- a/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
++++ b/protocol/src/main/java/net/md_5/bungee/protocol/Varint21FrameDecoder.java
+@@ -1,65 +1,78 @@
++/*
++ * Copyright (C) 2018 Velocity Contributors
++ *
++ * This program is free software: you can redistribute it and/or modify
++ * it under the terms of the GNU General Public License as published by
++ * the Free Software Foundation, either version 3 of the License, or
++ * (at your option) any later version.
++ *
++ * This program is distributed in the hope that it will be useful,
++ * but WITHOUT ANY WARRANTY; without even the implied warranty of
++ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
++ * GNU General Public License for more details.
++ *
++ * You should have received a copy of the GNU General Public License
++ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
++ */
++
+ package net.md_5.bungee.protocol;
+ 
++import net.md_5.bungee.protocol.VarintByteDecoder.DecodeResult;
+ import io.netty.buffer.ByteBuf;
+-import io.netty.buffer.Unpooled;
+ import io.netty.channel.ChannelHandlerContext;
+ import io.netty.handler.codec.ByteToMessageDecoder;
+-import io.netty.handler.codec.CorruptedFrameException;
++
+ import java.util.List;
+ 
+-public class Varint21FrameDecoder extends ByteToMessageDecoder
+-{
++public class Varint21FrameDecoder extends ByteToMessageDecoder {
+ 
+-    private static boolean DIRECT_WARNING;
++    private static final FastDecoderException BAD_LENGTH_CACHED =
++            new FastDecoderException("Bad packet length");
++    private static final FastDecoderException VARINT_BIG_CACHED =
++            new FastDecoderException("VarInt too big");
+ 
+     @Override
+-    protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception
+-    {
+-        // If we decode an invalid packet and an exception is thrown (thus triggering a close of the connection),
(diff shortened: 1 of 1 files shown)


Janmm14:
Better in terms of what?
Why create unneccessary additional object?
Did you create and run benchmarks?

Additionally there is a licensing problem, as in Waterfall is MIT while the code you want to insert here is GPL.
Or is that code in Velocity written by you originally?

xism4:
> Better in terms of what? Why create unneccessary additional object? Did you create and run benchmarks?
> 
> Additionally there is a licensing problem, as in Waterfall is MIT while the code you want to insert here is GPL. Or is that code in Velocity written by you originally?

I was talking to a dev from velocity, and he told me that if I followed the license, yes i could open this pull request.

Additionally I have done a couple of tests and it has been a little faster with the new patch.
Above all, a much cleaner source

Janmm14:
> > Better in terms of what? Why create unneccessary additional object? Did you create and run benchmarks?
> > Additionally there is a licensing problem, as in Waterfall is MIT while the code you want to insert here is GPL. Or is that code in Velocity written by you originally?
> 
> I was talking to a dev from velocity, and he told me that if I followed the license, yes i could open this pull request.
> 
> Additionally I have done a couple of tests and it has been a little faster with the new patch. Above all, a much cleaner source

You did not handle the GPL license correctly.

For inclusion of GPL code in a project, the complete project has to be released under GPL license.

This would mean that Waterfall has to change its license from MIT to GPL and may not be published under MIT license anymore. 

You should talk with a maintainer of Waterfall before you do so. As I am not sure whether a change to a far more restrictive license for Waterfall is desired here.

xism4:
> This would mean that Waterfall has to change its license from MIT to GPL and may not be published under MIT license anymore.

I talked with Five and i understond this, but ill wait @electronicboy, and let's see.

Kamillaova:
>allocating new object on every decode, for what?
image

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
```diff
@@ -0,0 +1,213 @@
...
+-        {
+-            in.skipBytes( in.readableBytes() );
++    protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
++        if (!ctx.channel().isActive()) {
++            in.clear();
+             return;
+         }
+ 
+-        in.markReaderIndex();
++        final VarintByteDecoder reader = new VarintByteDecoder();
```
Kamillaova, line 55: allocating object on every decode, for what?

xism4: > 

https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftVarintFrameDecoder.java#L41 (Velocity does same)

I have not wanted to modify the frame decoder, until there is a solid answer from the Waterfall developers, I should not change anything.

xIsm4:
> Better in terms of what? Why create unneccessary additional object? Did you create and run benchmarks?
> 
> Additionally there is a licensing problem, as in Waterfall is MIT while the code you want to insert here is GPL. Or is that code in Velocity written by you originally?

And apologies for not fully responding to you, I think I've fixed the license thing. 
This patch apart from having a much better code, has a system called (Trace-less) basically caching exceptions to prevent illegitimate connections and so on from causing a real performance problem compared to Waterfall.

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
```diff
@@ -0,0 +1,188 @@
...
+ import io.netty.buffer.ByteBuf;
+-import io.netty.buffer.Unpooled;
+ import io.netty.channel.ChannelHandlerContext;
+ import io.netty.handler.codec.ByteToMessageDecoder;
+-import io.netty.handler.codec.CorruptedFrameException;
++
+ import java.util.List;
+ 
+-public class Varint21FrameDecoder extends ByteToMessageDecoder
+-{
```
Kamillaova, line 29: https://github.com/PaperMC/Waterfall/blob/master/CONTRIBUTING.md

image

What is the reason for changing code style?

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
```diff
@@ -0,0 +1,188 @@
...
+-        {
+-            in.skipBytes( in.readableBytes() );
++    protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
++        if (!ctx.channel().isActive()) {
++            in.clear();
+             return;
+         }
+ 
+-        in.markReaderIndex();
++        final VarintByteDecoder reader = new VarintByteDecoder();
```
Kamillaova, line 55: Again. I suggests replace it with something like

```java
int readVarint;
int bytesRead;
DecodeResult result = DecodeResult.TOO_SHORT;

var start = in.readerIndex;
var end = in.writerIndex() + in.getLength();

for (; start < end; ++start) {
  var byte = in.getByte(start); // yes, rangecheck ;<

  if (k == 0 && bytesRead == 0) {
    // tentatively say it's invalid, but there's a possibility of redemption
    result = DecodeResult.RUN_OF_ZEROES;
    continue;
  }
  if (result == DecodeResult.RUN_OF_ZEROES) {
    break;
  }
  readVarint |= (k & 0x7F) << bytesRead++ * 7;
  if (bytesRead > 3) {
    result = DecodeResult.TOO_BIG;
    break;
  }
  if ((k & 0x80) != 128) {
    result = DecodeResult.SUCCESS;
    break;
  }
  continue;
}
```

Kamillaova (changes requested):

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
xism4: I followed imself Velocity's one, because it's simply easier to read and clean

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
xism4: Sure, i can modify it.

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
Kamillaova: > I followed imself Velocity's one, because it's simply easier to read and clean

Everything would be fine if in Java it was possible to allocate an object on the stack

Janmm14: This is an unneccessary change in a patch, the incentive is to keep patches simple so there is less merge conflict fixing needed.
This also means no import reordering or adding empty lines inbetween imports (ik its automatic by IDE usually).

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
```diff
@@ -0,0 +1,188 @@
...
+ import io.netty.handler.codec.ByteToMessageDecoder;
+-import io.netty.handler.codec.CorruptedFrameException;
++
+ import java.util.List;
+ 
+-public class Varint21FrameDecoder extends ByteToMessageDecoder
+-{
++public class Varint21FrameDecoder extends ByteToMessageDecoder {
+ 
+-    private static boolean DIRECT_WARNING;
```
Janmm14, line 32: The patch [0044-Don-t-use-a-bytebuf-for-packet-decoding.patch](https://github.com/PaperMC/Waterfall/blob/85c0a35f0b40a12ff0d7a35c6ea31287eecab707/BungeeCord-Patches/0044-Don-t-use-a-bytebuf-for-packet-decoding.patch) should be edited if you want to remove this line.

Janmm14 (changes requested):

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
Janmm14: Sounds kinda useless to use the DecodeResult at all. Instead of breaking and later switching over the enum, we should simply program the actions (exception throw) we'd take on each result immediately.

The if (run of zeroes) after the continue is also never being used.

And without you changing this, i woudln't mark this as resolved.

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
xism4: > This is an unneccessary change in a patch, the incentive is to keep patches simple so there is less merge conflict fixing needed. This also means no import reordering or adding empty lines inbetween imports (ik its automatic by IDE usually).

I will follow that, thanks

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
xism4: Got it

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
ghost: Actually on HotSpot JVM this is optimised

astei: I specifically crafted `VarintByteDecoder` in such a fashion so that the HotSpot JVM would optimize away the object creation through inlining the code in `AbstractByteBuf.forEachByte()`, allowing escape analysis to do the rest of the work.

Appearances can be quite deceiving!

xism4: > Appearances can be quite deceiving!

I will left it like this then

Janmm14: > I specifically crafted `VarintByteDecoder` in such a fashion so that the HotSpot JVM would optimize away the object creation through inlining the code in `AbstractByteBuf.forEachByte()`, allowing escape analysis to do the rest of the work.
> 
> Appearances can be quite deceiving!

Inmy opinion its unneccessary fancyness youre doing there.

astei: It's not fanciness, it was me observing the behavior of the JVM and guiding it in the best direction in optimizing the code.

Janmm14: So you say the overall shorter and in my opinion equally readable while loop is not getting to the same optimizations?

BungeeCord-Patches/0066-Use-Velocity-VarintFrameDecoder.patch
```diff
@@ -0,0 +1,188 @@
...
+diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/VarintByteDecoder.java b/protocol/src/main/java/net/md_5/bungee/protocol/VarintByteDecoder.java
+new file mode 100644
+index 00000000..2e4ccf05
+--- /dev/null
++++ b/protocol/src/main/java/net/md_5/bungee/protocol/VarintByteDecoder.java
+@@ -0,0 +1,68 @@
++/*
++ * Courtesy of Velocity
++ * See https://github.com/VelocityPowered/Velocity/commit/5ceac16a821ea35572ff11412ace8929fd06e278
++ */
```
Janmm14, line 133: I see, this commit is before Velocity's implementation changed to GPL.

xism4: > I see, this commit is before Velocity's implementation changed to GPL.

Yes, i saw even on other patches of Waterfall same

xIsm4:
I will close this due multiple opinions with really diferent points.

PR #799, opened 2023-04-03
Use ByteBuf#copy only when entity rewrite is actually used

BoomEaro:
This pull request returns the ability to slice ByteBuf when decoding a packet in the MinecraftDecoder class as it did before, but only if entity rewrite was never used.
If I understand correctly, the original reason for using copy is that entity rewrite can write more bytes than is possible in the slice itself.
Therefore, most likely there is no point in constantly copying bytes even when we are not going to change them.

I would like to know your opinion on this, what it could possibly break and whether it makes any sense at all, since I have not had much experience with netty yet.

### DIFF
--- a/BungeeCord-Patches/0067-Always-slice-until-entity-rewrite-is-being-used-in-M.patch
+++ b/BungeeCord-Patches/0067-Always-slice-until-entity-rewrite-is-being-used-in-M.patch
@@ -1,0 +1,57 @@
+From fa7dcf4b0fa0b5aaf20c9071b547699ae43ab530 Mon Sep 17 00:00:00 2001
+From: BoomEaro <21033866+BoomEaro@users.noreply.github.com>
+Date: Mon, 3 Apr 2023 13:03:53 +0300
+Subject: [PATCH] Always slice until entity rewrite is being used in
+ MinecraftDecoder
+
+
+diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+index ac83e325..b6950ff1 100644
+--- a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
++++ b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+@@ -20,6 +20,8 @@ public class MinecraftDecoder extends MessageToMessageDecoder<ByteBuf>
+     private int protocolVersion;
+     @Setter
+     private boolean supportsForge = false;
++    @Setter
++    private boolean slice = true; // Waterfall
+ 
+     public MinecraftDecoder(Protocol protocol, boolean server, int protocolVersion) {
+         this.protocol = protocol;
+@@ -38,7 +40,7 @@ public class MinecraftDecoder extends MessageToMessageDecoder<ByteBuf>
+         }
+ 
+         Protocol.DirectionData prot = ( server ) ? protocol.TO_SERVER : protocol.TO_CLIENT;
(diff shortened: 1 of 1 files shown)


electronicboy:
This was long on the list of stuff that I wanted to look into, given that we're not rewriting this stuff, I did wanna look into adding a mechanism to literally just skip packets and let the proxy not care about them, even if registered

Janmm14:
There's no such thing as skipping packets, you always have to check packet id. And slicing instead of copying is already a big progress.
Not sure what else can be done here to optimize performance.

electronicboy:
I mean skip the decode + handle + encode chain for unneeded packets in certain configs; if we're not rewriting entity metadata, then we can skip a bunch of work

BungeeCord-Patches/0067-Always-slice-until-entity-rewrite-is-being-used-in-M.patch
```diff
@@ -0,0 +1,57 @@
...
+diff --git a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+index ac83e325..b6950ff1 100644
+--- a/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
++++ b/protocol/src/main/java/net/md_5/bungee/protocol/MinecraftDecoder.java
+@@ -20,6 +20,8 @@ public class MinecraftDecoder extends MessageToMessageDecoder<ByteBuf>
+     private int protocolVersion;
+     @Setter
+     private boolean supportsForge = false;
++    @Setter
++    private boolean slice = true; // Waterfall
```
Janmm14, line 17: Shouldn't we default to the safe side, copy instead of slice?

Janmm14:
> I mean skip the decode + handle + encode chain for unneeded packets in certain configs; if we're not rewriting entity metadata, then we can skip a bunch of work

Bungee doesn't use decoding into object fields for most entity rewriting stuff.

electronicboy:
True, but part of it was for other stuff like scoreboards, etc; it's probably not a big deal, and probs not something I'm ever going to care to get to considering the state of this project

BungeeCord-Patches/0067-Always-slice-until-entity-rewrite-is-being-used-in-M.patch
BoomEaro: > Shouldn't we default to the safe side, copy instead of slice?

As far as I remember, the only situation where the proxy actually changes the buffer is when an entity map instance appears in the UserConnection.
Before the proxy creates the UserConnection, the proxy decodes the Handshake, Login, and a couple of other packets. (when the state is not yet FINISHING in InitialHandler)
It doesn't seem to make sense at this stage specifically to copy the buffer for these packets.

Janmm14: btw why you don't set this directly at initialization of minecraftdecoder to the config value of rewriting disabled?

if we don't want to clone packets given the game phase, this should somewhere be explicitely mentioned - either in code or a simple comment

bob7l:
This commit would likely break several plugins like ViaVersion that are expecting a grow-able, copied buffer. If anything, it should be made into a configurable option for the user to decide.

Janmm14:
> This commit would likely break several plugins like ViaVersion that are expecting a grow-able, copied buffer. If anything, it should be made into a configurable option for the user to decide.

Statement from someone knowing viaverson-bungee is needed. Don't want some useless config option or startup flag.

xism4 (approved):

xism4:
> 

This would only theoretically occur if the option were disabled

PR #804, opened 2023-06-03
Updated Upstream (BungeeCord)

linsaftw:


electronicboy:
Not exactly fond of exclusions, really just need to get tca updated at some point

The other one looks like some library just needs updating on our side, would need to look when at my PC

electronicboy (approved):

electronicboy:
Thanks, slf4j is apparently in a weird borked state, and I'll try and speak to minecrell at some point

PR #807, opened 2023-07-08
Add forwarding system

xism4:
This pull request comes from the other one not accepted due to some factors that were fixed here

* Added 1.19 and 1.20 support
*  Cut down excesive logg-in
* Tested all the modes (Velocity_forwarding with modern backends, Legacy one and BungeeGuard)

@electronicboy 

You can read what this adds on https://github.com/PaperMC/Waterfall/pull/663


xism4:
Ignore patch error, it's just a editor issue because im not that friendly with patches.

PR #818, opened 2023-09-24
some tweaks

l3nnartt:
update of actions because node12 is deprecated
use more java lts versions to build
update readme version from waterfall


.github/workflows/build.yml
```diff
@@ -12,12 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [17,8]
+        java: [8, 11, 17]
```
Janmm14: 21

l3nnartt: i tried that;
https://github.com/l3nnartt/Waterfall/actions/runs/6289342348/job/17075748849

BlockyTheDev: The problem here is that `distribution: 'adopt'` isn't working anymore for Java 21, because the movement from AdopOpenJDK to Eclipse Temurin there aren't JDK 21 release distributions of AdoptOpenJDK, ... Because of that the distribution needs to be changed to `temurin`.

l3nnartt: youre right, thanks!

l3nnartt: Looks like its not available yet, [zulu](https://github.com/actions/setup-java?tab=readme-ov-file#supported-distributions) already has 21, but we can also wait for temurin ig

.github/workflows/build.yml
```diff
@@ -15,9 +15,9 @@ jobs:
...
       fail-fast: true
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Install Java ${{ matrix.java }}
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java }}
           distribution: 'adopt'
```
Kichura: "adopt" can be updated to "temurin".

BlockyTheDev: Already changed in this pull request as far as I see. Waiting on Temurin 21 release is necessary...

MiniDigger: which is waiting because of some stupid legal bullshit, maybe it should be changed to a vendor that actually releases timely, since this isnt the first case where the eclipse foundation is slow.

BlockyTheDev: > which is waiting because of some stupid legal bullshit, maybe it should be changed to a vendor that actually releases timely, since this isnt the first case where the eclipse foundation is slow.

Yes, the Eclipse foundation is very slow. I think vendors like Azul and the others had to do the same legal things and got it done on time. Maybe switch to Azul Zulu?

l3nnartt: done

BlockyTheDev (approved):

xism4 (approved):

.github/workflows/build.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java }}
-          distribution: 'temurin'
+          distribution: 'zulu'
```
Kichura, line 23: Using zulu isn't recommended for CI due to certain tweaks being worse than temurin, Also consider including `check-latest: true` below distribution.

Kichura (changes requested):

.github/workflows/build.yml
BlockyTheDev: Never heard of that about `zulu`. Do you have any links for me, where I can read about it, because I'm interested about the background of the named tweaks?

I suggested `zulu` because I'm using it myself for a long time now and it is used by many bigger projects for CI like `netty`, `lombok` and `guava`...

Janmm14: zulu was previously chosen due to faster update times. In general I think we dont neccessarily need the very latest version of the jdk, testing the build and tests with multiple jdks is just there to ensure basic compatibility with the LTS versions. The tests of bungee have no capability of detecting bungee-only incompatibilities with a minor update of the jdk, if bungee tests fail due to minor jdk update, the jdk update is faulty.

Janmm14 (approved):

electronicboy (approved):

PR #825, opened 2023-10-17
Add tcp fast open

xism4:
It was implemented because Velocity comes with this system, and it should speed up the opening of connections between server-backend at level 3 for systems that support Netty's epoll

Minecraft clients support this feature.

https://en.wikipedia.org/wiki/TCP_Fast_Open

### DIFF
--- a/BungeeCord-Patches/0067-Add-tcp-fast-open.patch
+++ b/BungeeCord-Patches/0067-Add-tcp-fast-open.patch
@@ -1,0 +1,21 @@
+From 975be85816950e37799375a99fd69d089cee6347 Mon Sep 17 00:00:00 2001
+From: Ismael Hanbel <soportexism4@gmail.com>
+Date: Tue, 17 Oct 2023 17:01:57 +0200
+Subject: [PATCH] Add tcp fast open
+
+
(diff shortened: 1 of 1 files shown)


BoomEaro:
I decided to look into this more and came to the conclusion that it makes no sense:

1. The ChannelOption.TCP_FASTOPEN constant has nothing to do with setting the TFO mode.
The value set there configures the maximum number of outstanding (waiting to be accepted) TFO connections. (https://netty.io/4.1/api/io/netty/channel/ChannelOption.html#TCP_FASTOPEN)
2. The game client never sets the TCP_FASTOPEN_CONNECT constant to enable TFO. (proxy also should do this for downstream to enable this feature)
3. Spigot and paper does not support TFO by default.
4. TFO mode should be configured only here on linux systems: /proc/net/ipv4/tcp_fastopen
5. Client and server should support this at the system level and at the application level. Windows does not support this (most minecraft users are windows users)

It’s just not clear to me why velocity has this support, while neither paper nor windows users will be able to use it. In addition, for TFO to work, you need to explicitly configure the BOTH mode (for proxy) in the system configuration. (by default enabled only CLIENT mode)

xism4:
I will close this since this needs backend impl for Spigot and forks

PR #838, opened 2023-12-23
Updated Upstreaem (BungeCord)

linsaftw:


electronicboy:
I apparently failed to notice this, but this is generally unmergable without the correct commit headers for an upstream, see the scripts folder

linsaftw:
Thx

PR #839, opened 2024-01-04
Add config with http request proxy

lintx:
Add settings to use the optional proxy option when the server requests the mojang server. In some areas, the mojang website is often unable to be accessed. This is necessary.


MiniDigger:
why would the software need to support this? have you tried just setting the proxy via system props?
`java -Dhttp.proxyHost=yourproxy.com -Dhttp.proxyPort=8080 -jar waterfall.jar`

lintx:
> why would the software need to support this? have you tried just setting the proxy via system props? `java -Dhttp.proxyHost=yourproxy.com -Dhttp.proxyPort=8080 -jar waterfall.jar`

BungeeCord uses netty as the http request library, and netty does not support setting the proxy using jvm parameters.

PR #859, opened 2025-03-25
Updated Upstream (from ArkFlame's Waterfall)

linsaftw:
Updated Upstream to support 1.21.5 Minecraft


linsaftw:
Something is wrong with this one... BungeeCord removed the plugin message handling but this is causing issues. I will revert their change because it causes issues with plugins that use plugin message.

